PR #1, opened 2025-02-05
Add typing of values

N1ark:
Allow `Value.t` and `Value.ty` to carry type information; add `Value.sbool` to represent the type of symbolic booleans.
Update `bfa_c` accordingly, by making `Csymex` use `Typed` rather than `SValue`.

### DIFF
--- a/bfa/lib/symex/plist.ml
+++ b/bfa/lib/symex/plist.ml
@@ -11,12 +11,12 @@ module type SInt_sig = sig
   include Stdlib.Map.OrderedType with type t := t
 
   val pp : t Fmt.t
-  val sem_eq : t -> t -> Value.t
+  val sem_eq : t -> t -> Value.sbool Value.t
   val of_int : int -> t
-  val in_range : t -> t * t -> Value.t
-  val greater_or_equal : t -> t -> Value.t
+  val in_range : t -> t * t -> Value.sbool Value.t
+  val greater_or_equal : t -> t -> Value.sbool Value.t
   val subst : (Var.t -> Var.t) -> t -> t
-  val iter_vars : t -> (Var.t * Value.ty -> unit) -> unit
+  val iter_vars : t -> (Var.t * 'a Value.ty -> unit) -> unit
 end
 
 module Make (Symex : Symex.S) (SInt : SInt_sig with module Symex = Symex) =
--- a/bfa/lib/symex/plist.ml
+++ b/bfa/lib/symex/plist.ml
@@ -10,11 +10,13 @@ module type SInt_sig = sig
 
   include Stdlib.Map.OrderedType with type t := t
 
+  type sbool_v := Value.sbool Value.t
+
   val pp : t Fmt.t
-  val sem_eq : t -> t -> Value.sbool Value.t
+  val sem_eq : t -> t -> sbool_v
   val of_int : int -> t
-  val in_range : t -> t * t -> Value.sbool Value.t
-  val greater_or_equal : t -> t -> Value.sbool Value.t
+  val in_range : t -> t * t -> sbool_v
+  val greater_or_equal : t -> t -> sbool_v
   val subst : (Var.t -> Var.t) -> t -> t
   val iter_vars : t -> (Var.t * 'a Value.ty -> unit) -> unit
 end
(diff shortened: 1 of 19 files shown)


bfa/lib/symex/plist.ml
```diff
@@ -11,12 +11,12 @@ module type SInt_sig = sig
   include Stdlib.Map.OrderedType with type t := t
 
   val pp : t Fmt.t
-  val sem_eq : t -> t -> Value.t
+  val sem_eq : t -> t -> Value.sbool Value.t
```
giltho, line 14: Hmm, should probably have some
```ocaml
module Value_ty_aliases (V : Values) = struct type boolv = Value.sbool Value.t end
```
and then
```
open Value_ty_aliases(Values)
```
in several places to avoid having too many of these longs types

bfa/lib/symex/symex.ml
```diff
@@ -4,15 +4,19 @@ module type Base = sig
   module Value : Value.S
   module MONAD : Monad.Base
 
-  val assume : Value.t list -> unit MONAD.t
+  type 'a v := 'a Value.t
+  type 'a vt := 'a Value.ty
+  type sbool := Value.sbool
```
giltho, line 9: Yeah that also works, maybe some mix of both?

bfa_c/lib/csymex.ml
```diff
@@ -1,6 +1,15 @@
 module SYMEX = Bfa_symex.Symex.Make_iter (Z3solver)
 include SYMEX
 
+let check_nonzero (t : Typed.T.sint Typed.t) :
+    ([> Typed.T.nonzero ] Typed.t, [> `NonZeroIsZero ], 'fix) Result.t =
+  let open Syntax in
+  let open Typed.Infix in
+  if%sat t ==@ Typed.zero then Result.error `NonZeroIsZero
+  else Result.ok (Typed.cast t)
```
giltho, line 9: Why has this moved here?

bfa_c/lib/heap.ml
```diff
@@ -15,26 +15,13 @@ let with_error_loc_as_call_trace () f =
...
-      | Some f -> Some (fun loc -> f (Typed.type_ loc))
-    in
-    let+ v = Csymex.nondet ?constrs TLoc in
-    Typed.type_ v
+  type t = T.sloc Typed.t
 
-  let iter_vars = Typed.iter_vars
+  let pp = ppa
+  let fresh ?constrs () = Csymex.nondet ?constrs Typed.t_loc
 end)
```
giltho, line 25: So much cleaner!

bfa_c/lib/typed.mli
```diff
@@ -38,41 +36,40 @@ val t_seq : ([< any ] as 'a) ty -> [> 'a sseq ] ty
...
+val pp : 'a Fmt.t -> 'a t Fmt.t
+val ppa : 'a t Fmt.t
 val equal : ([< any ] as 'a) t -> 'a t -> bool
 val compare : ([< any ] as 'a) t -> 'a t -> int
 
 (** Typed constructors *)
 
-val sem_eq : ([< any ] as 'a) t -> 'a t -> [> sbool ] t
+val sem_eq : 'a t -> 'b t -> sbool t
 val sem_eq_untyped : 'a t -> 'b t -> [> sbool ] t
```
giltho, line 63: I think I really want `sem_eq` to receive two values of the same type? Or am I missing something here?

bfa_c/lib/csymex.ml
N1ark: Used to be in `Typed`, which is imported in `Z3Solver`, which is imported in `Csymex`, which is used in the snippet, causing a circular dependency
I had to remove all functions/modules refering to `Csymex` in `Typed`, which turned out quite simple now that everything is typed properly

bfa/lib/symex/plist.ml
N1ark: Yeah I'll add `type boolv := ...` for the few cases where it's annoying, you're right
I think the functor is a bit over of the top?

bfa_c/lib/typed.mli
N1ark: I thought about it and wasn't sure! To me it made sense that one might one to compare two arbitrary values for equality, eg. a `cint` and a `ptr`? Not sure if disallowing this breaks anything, will try

N1ark: Ok yeah that breaks nothing nice

giltho (approved):

PR #10, opened 2025-03-27
More work on bfa-c

giltho:
Opening the PR early so I can check that my update of cerberus worked in CI

### DIFF
--- a/bfa/lib/bfa_std/ordered_type.ml
+++ b/bfa/lib/bfa_std/ordered_type.ml
@@ -1,0 +1,5 @@
+module type S = sig
+  include Stdlib.Map.OrderedType
+
+  val pp : Format.formatter -> t -> unit
+end
--- a/bfa_c/lib/heap.ml
+++ b/bfa_c/lib/heap.ml
@@ -220,3 +220,33 @@ let consume (serialized : serialized) (st : t) :
   | Missing fixes ->
       let fixes = List.map (fun fix -> { heap = fix; globs = [] }) fixes in
       Missing fixes
+
+let get_global (sym : Cerb_frontend.Symbol.sym) (st : t) =
+  let+ loc, globs = Globs.get sym st.globs in
+  let ptr = Typed.Ptr.mk loc 0s in
+  (ptr, { st with globs })
+
+let init_prog_state (prog : Ail_tys.sigma) : t Csymex.t =
+  let produce_zero (ptr : [< T.sptr ] Typed.t) ty (state : t) : t Csymex.t =
+    let loc = Typed.Ptr.loc ptr in
+    let offset = Typed.Ptr.ofs ptr in
+    let* len = Layout.size_of_s ty in
+    let serialized : serialized =
+      {
+        heap = [ (loc, Freeable.Alive [ Tree_block.Zeros { offset; len } ]) ];
+        globs = [];
+      }
+    in
+    produce serialized state
+  in
+  Csymex.fold_list prog.declarations ~init:empty ~f:(fun (state : t) decl ->
+      let id, (_loc, _attrs, decl) = decl in
+      match decl with
+      | Cerb_frontend.AilSyntax.Decl_function _ -> Csymex.return state
+      | Decl_object (_, _, _, ty) ->
+          (* TODO: handle other parameters here *)
+          let* loc, globs = Globs.get id state.Heap_intf.Template.globs in
+          let state = { state with globs } in
+          let ptr = Typed.Ptr.mk loc 0s in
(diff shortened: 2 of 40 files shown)


bfa/lib/bfa_std/ordered_type.ml
```diff
@@ -0,0 +1,5 @@
+module type S = sig
+  include Stdlib.Map.OrderedType
+
+  val pp : Format.formatter -> t -> unit
```
N1ark, line 4: note: maybe decide if we use `Format.formatter -> t -> unit` or `t Fmt.t`

bfa_c/lib/heap.ml
```diff
@@ -65,6 +81,25 @@ let log action ptr st =
...
 
+let with_heap st f =
+  let+ res = f st.heap in
+  match res with
+  | Bfa_symex.Compo_res.Ok (v, h) ->
+      Bfa_symex.Compo_res.Ok (v, { st with heap = h })
+  | Missing fixes ->
+      let fixes = List.map (fun fix -> { heap = fix; globs = [] }) fixes in
+      Missing fixes
+  | Error e -> Error e
```
N1ark, line 92: Nice, stealing this :p

bfa_c/lib/heap.ml
```diff
@@ -134,23 +172,81 @@ let free (ptr : [< T.sptr ] Typed.t) (st : t) :
...
+        heap = [ (loc, Freeable.Alive [ Tree_block.Zeros { offset; len } ]) ];
+        globs = [];
+      }
+    in
+    produce serialized state
+  in
+  Csymex.fold_list prog.declarations ~init:empty ~f:(fun (state : t) decl ->
+      let id, (_loc, _attrs, decl) = decl in
+      match decl with
+      | Cerb_frontend.AilSyntax.Decl_function _ -> Csymex.return state
```
N1ark, line 245: Would this be where you manage function pointers? 👀

bfa_c/lib/svalue.ml
```diff
@@ -178,14 +178,20 @@ let and_ v1 v2 =
...
+    | Bool b1, Bool b2 -> bool (b1 = b2)
+    | Ptr (l1, o1), Ptr (l2, o2) -> and_ (sem_eq l1 l2) (sem_eq o1 o2)
+    | Binop (Plus, v1, v2), Binop (Plus, v3, v4) when equal v1 v3 ->
+        sem_eq v2 v4
+    | _ ->
+        (* We put equality in some sort of normal form where 
+         element with the smallest id is on the LHS. *)
+        (* TODO: Generalise to all commutative operators. *)
+        if v1.tag <= v2.tag then Binop (Eq, v1, v2) <| TBool
+        else Binop (Eq, v2, v1) <| TBool
```
N1ark, line 194: minor but could make a `mk_binop` fn to do just this for all :)

N1ark (approved):
lgtm :)

bfa/lib/bfa_std/ordered_type.ml
giltho: I tend to use `Format.formatter -> t -> unit` where I think about it to reduce dependency on Fmt, and make doc more readable for people who don't use Fmt. I should maybe uniformise

giltho: Any opinion btw?

giltho: I changed it the stdlib version
I didn't touch bfa_rust, but you can easily do it:

search: `(('|\w|\s)+) Fmt.t`
replace: ` Format.formatter -> $1 -> unit`

Not perfect but worked for all the cases for me

bfa_c/lib/heap.ml
giltho: Yep

bfa_c/lib/svalue.ml
giltho: Though at some point we might want to have better normal forms.
For instance, in additive operations, we might want to put all concrete operands e.g. on the right-hand side.
At some point, that means that reductions can work more efficiently

bfa/lib/bfa_std/ordered_type.ml
N1ark: Ahhh fair enough! I didn't consider that, I just use Fmt everywhere :) Will do

N1ark: I personally prefer `'a Fmt.t` because it's much shorter but I don't feel strongly about it so I'm happy with the full form

PR #12, opened 2025-03-29
Parametric pointers

N1ark:
The interpreter currently manipulates pointers directly -- this is possible because we are in monomorphised code, where all field offsets are known. To enable Ruxt, that runs on polymorphic code, we must abstract this away.

Added a module `Sptr` that does just this -- it abstracts a pointer and its metadata (for fat pointers). Moved the current pointer arithmetic logic to a module `ArithPtr`. The interpreter and STD functions are now parametric on the pointer module, that is provided by the heap module.

The logic for converting to and from C values to rust_val has also been moved to an `encoder.ml` file, to allow `Layout.ml` to work irrespective of the pointer type (while `encoder.ml` assumes `ArithPtr` is used).

### DIFF
--- a/bfa_rust/lib/heap.ml
+++ b/bfa_rust/lib/heap.ml
@@ -3,6 +3,7 @@ open Typed.Infix
 open Typed.Syntax
 module T = Typed.T
 open Rustsymex
+module Sptr = Sptr.T
 
 type 'a err = 'a * Call_trace.t
 
@@ -41,7 +42,7 @@ end
 
 type t = {
   heap : Tree_block.t Freeable.t SPmap.t option;
-  globals : Charon_util.rust_val GlobMap.t;
+  globals : Sptr.t GlobMap.t;
 }
 [@@deriving show { with_path = false }]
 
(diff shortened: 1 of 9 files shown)


bfa_rust/lib/sptr.ml
```diff
@@ -0,0 +1,90 @@
...
+
+  (** If this is the null pointer *)
+  val is_null : t -> sbool Typed.t
+
+  (** If these two pointers are at the same location (ie. same allocation) *)
+  val is_same_loc : t -> t -> sbool Typed.t
+
+  (** The distance, in bytes, between two pointers -- assumes they are at the
+      same location. *)
+  val distance : t -> t -> sint Typed.t
```
giltho, line 24: It's possible that this function (and maybe others?) should return a `symex` instead of an integer.
For some Pedro cases, we might not be able to compute this, and we'll just give-up

giltho (commented):
Only real comment to address is the null_loc one

giltho:
Very very sad

bfa_rust/lib/heap.ml
```diff
@@ -164,7 +167,7 @@ let copy_nonoverlapping ~dst ~src ~size st =
   let open Typed.Infix in
   let@ () = with_error_loc_as_call_trace () in
   let@ () = with_loc_err () in
-  if%sat Typed.Ptr.is_at_null_loc dst ||@ Typed.Ptr.is_at_null_loc src then
+  if%sat Sptr.is_null dst ||@ Sptr.is_null src then
```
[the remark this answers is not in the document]
N1ark: Ah right, I actually just didn't notice the change :p you're right though, that's the right check -- I updated Sptr.is_null.

bfa_rust/lib/sptr.ml
N1ark: yes, makes sense -- for now I'll leave it as is, but I'll keep it in mind!

PR #14, opened 2025-04-03
Tree borrows

N1ark:
- Add tree borrows to Rusteria:
  - The heap store, for each location, the tree block along with the tree borrow, ie. the tree structure of tags.
  - Within the tree block leaves, along with the stored value is a mapping of tag to state; on reads and writes, the state of all tags is updated and checked for UB.
- Make tree blocks/`encoder.ml` use `rust_val`, rather than `T.cval Typed.t` -- this is needed to allow storing ghost state (e.g. the tag of pointers), and is a first step towards "best effort abstraction", where non-atomic values (e.g. structs) are stored in treeblocks as a whole.
- Remove `meta` from `Sptr` -- the value of pointers is now defined as `Ptr of 'ptr_ty full_ptr`, where `'ptr_ty` is the parametric pointer type, and `'ptr_ty full_ptr` is `ptr_ty * T.cval Typed.t option`. This allows meta to be accessed and modified regardless of the type of pointers.
- Improved the packaging of the Kani lib -- thanks to `dune-sites`, the Rust lib comes with Rusteria, and is re-compiled when needed.
- Add tests for tree borrows
- Download Rust and Charon in CI
- Add some reductions (for <=, <, ==, %, rem)


bfa_rust/lib/heap.ml
```diff
@@ -84,160 +92,167 @@ let log action ptr st =
...
-    Result.error `NullDereference
-  else
-    let** tree_to_write =
-      with_ptr_read_only src st (fun ~ofs block ->
-          let++ tree, _ = Tree_block.get_raw_tree_owned ofs size block in
-          tree)
-    in
-    with_ptr dst st (fun ~ofs block ->
-        Tree_block.put_raw_tree ofs tree_to_write block)
+  let** tree_to_write, st =
```
giltho, line 194: You should somehow check non-overlappingness

bfa_rust/lib/heap.ml
```diff
@@ -253,3 +268,52 @@ let load_str_global str ({ globals; _ } as st) =
 let load_global g ({ globals; _ } as st) =
   let ptr = GlobMap.find_opt (Global g) globals in
   Result.ok (ptr, st)
+
+let borrow (((raw_ptr, parent) as ptr), meta)
```
giltho, line 272: Hehehehehe exciting

bfa_rust/lib/interp.ml
```diff
@@ -569,30 +579,55 @@ module Make (Heap : Heap_intf.S) = struct
...
+                    list ~sep:comma
+                    @@ pair ~sep:(any "->") string UllbcAst.pp_block_id)
+                  options UllbcAst.pp_block_id default pp_rust_val discr);
+            let compare_discr =
+              match discr with
+              | Base discr -> fun (v, _) -> discr ==@ value_of_scalar v
+              | Ptr (ptr, _) ->
+                  fun (v, _) ->
+                    if Z.equal Z.zero v.value then Sptr.is_at_null_loc ptr
+                    else failwith "Can't compare pointer with non-0 scalar"
```
giltho, line 621: Not sure that's entirely true. I need to write down a document for handling pointers vs integers

bfa_rust/lib/layout.ml
```diff
@@ -404,3 +394,39 @@ let rec nondet : Types.ty -> 'a rust_val Rustsymex.t =
...
+      Fmt.failwith "to_zeros: unsupported literal type %a" Types.pp_literal_type
+        t
+
+let rec zeroed ~(null_ptr : 'a) : Types.ty -> 'a rust_val option = function
+  | TLiteral lit_ty -> ( try Some (Base (zeroed_lit lit_ty)) with _ -> None)
+  | TRawPtr _ -> Some (Ptr (null_ptr, None))
+  | TRef _ -> None
+  | TAdt (TTuple, { types; _ }) ->
+      Utils_.List_ex.join_options (zeroed ~null_ptr) types
+      |> Option.map (fun fields -> Tuple fields)
```
giltho, line 410: Yep, this is wrong because join_options reverse the list 😬

bfa_rust/lib/tree_block.ml
```diff
@@ -621,8 +669,9 @@ let uninit_range ofs size t =
 let subst_serialized subst_var (serialized : serialized) =
   let v_subst v = Typed.subst subst_var v in
   let subst_atom = function
-    | TypedVal { offset; ty; v } ->
-        TypedVal { offset = v_subst offset; ty; v = v_subst v }
+    | TypedVal _ ->
+        failwith "not_impl: subst_serialized TypedVal"
```
giltho, line 673: Why failwith instead of not_impl?

giltho: Ah, because subst isn't a symbolic process. It should prob be. But still, what's wrong with substituting typed vals?

bfa_rust/lib/tree_borrow.ml
```diff
@@ -1,54 +1,189 @@
...
+(* let[@tailrec] rec root n = match n.parent with None -> n | Some p -> root p *)
+
+(** [find n t] Looks for the node with tag [t] in the tree rooted at [n] *)
+let find n t =
   let[@tailrec] rec aux = function
-    | [] -> false
-    | n :: _ when n.tag == n2.tag -> true
+    | [] -> raise Not_found
+    | n :: _ when n.tag = t -> n
     | n :: ns -> aux (n.children @ ns)
```
giltho, line 93: might as well use `Iter` here, concatenating lists again and again is not a bit expensive

bfa_rust/lib/tree_borrow.ml
```diff
@@ -1,54 +1,189 @@
...
+  in
+  let root' = aux root in
+  if !found then root' else raise Not_found
+
+let all_nodes root =
+  let rec aux acc = function
+    | [] -> acc
+    | n :: ns -> aux (n :: acc) (n.children @ ns)
+  in
+  aux [] [ root ]
```
giltho, line 121: Let's return a sequence or an `Iter`. If it's ever converted to a list, use seq

.github/workflows/ci.yml
```diff
@@ -22,6 +22,17 @@ jobs:
       - name: Setup Z3
         uses: cda-tum/setup-z3@v1
         id: z3
+      - name: Setup Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+      - name: Setup Charon
+        run: |
+          brew install make
+          cd ..
+          git clone https://github.com/N1ark/charon.git
```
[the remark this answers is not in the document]
N1ark: Right now monomorphisation is in a place where it seems to work (for my needs, at least), so I never need to touch my branch, I think it's ok to leave as is for now

bfa_rust/lib/heap.ml
```diff
@@ -52,7 +54,10 @@ type serialized = Tree_block.serialized Freeable.serialized SPmap.serialized
 let serialize { heap; _ } =
   match heap with
   | None -> []
-  | Some heap -> SPmap.serialize (Freeable.serialize Tree_block.serialize) heap
```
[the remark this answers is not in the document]
N1ark: Yes I don't use it but it doesn't cost anything to keep it there for the future I guess

bfa_rust/lib/heap.ml
N1ark: I do it when calling it, in `std_funs.ml`:
```ocaml
        (* check for overlap *)
        let from_ptr_end = Sptr.offset from_ptr size in
        let to_ptr_end = Sptr.offset to_ptr size in
        if%sat
          Sptr.is_same_loc from_ptr to_ptr
          &&@ (Sptr.distance from_ptr to_ptr_end
              <@ 0s
              &&@ (Sptr.distance to_ptr from_ptr_end <@ 0s))
        then Heap.error (`StdErr "copy_nonoverlapping overlapped") state
        else Result.ok ()
```

giltho:
Ah, comment commits don't appear here but your reduction of `not` for the `Or` case is wrong

bfa_rust/lib/tree_block.ml
N1ark: Nothing it just requires me to implement subst for `rust_val` which I haven't gotten around to and don't need anyways :p

bfa/lib/bfa_std/utils.ml
```diff
@@ -12,6 +12,14 @@ module List_ex = struct
...
     if List.is_empty errors then Ok oks else Error errors
 
+  let join_options fn opts =
+    let[@tailrec] rec aux acc = function
+      | [] -> Some acc
+      | x :: xs -> (
+          match fn x with Some x -> aux (x :: acc) xs | None -> None)
+    in
+    aux [] opts
+
```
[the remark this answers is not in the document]
N1ark: Fixed!

bfa_rust/lib/tree_borrow.ml
N1ark: Done!

bfa_rust/lib/tree_borrow.ml
N1ark: Done!

N1ark:
> Ah, comment commits don't appear here but your reduction of `not` for the `Or` case is wrong

oops! fixed

bfa_rust/lib/heap.ml
```diff
@@ -40,9 +40,11 @@ module GlobMap = struct
     Fmt.pf fmt "%a" (Fmt.iter_bindings ~sep:Fmt.comma iter pp_pair) m
 end
 
+type sub_t = Tree_block.t * Tree_borrow.t
```
giltho, line 43: Why does the `_t` stand for? If it's for `type`, just call it `sub`. If it's for tree, I'd call it `sub_tree`.

bfa_rust/lib/heap.ml
```diff
@@ -40,9 +40,11 @@ module GlobMap = struct
...
+
 type t = {
-  heap : Tree_block.t Freeable.t SPmap.t option;
-  globals : Sptr.t GlobMap.t;
+  heap : (Tree_block.t * Tree_borrow.t) Freeable.t SPmap.t option;
```
giltho, line 46: Should that be `sub_t Freeable.t SPmap.t option`?

bfa_rust/lib/heap.ml
N1ark: It was for type :) will get rid of it

giltho: Saying this because in OCaml, types and values live in different namespaces so there is no confusion possible. So I tend to remove all `_t`

bfa_rust/lib/heap.ml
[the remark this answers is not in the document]
giltho: Yeah but like, at the moment it'd be unsound to use this `serialize` in spec generation because information is not serialised. So might be slightly better to remove it so we don't forget to properly implement it when needed

bfa_rust/lib/heap.ml
N1ark: it could be but then I'd need to define `pp_sub_t` etc which I wanted to avoid :p but really it's not that serious so will change

giltho: No fair enough, let's not expose it :p

N1ark: too late hahaha i better avoiding the duplication anyways

PR #18, opened 2025-04-11
fix symex state with seq

### LINKED ISSUE #8: Make_seq and Make_iter use fuel differently
Given the same fuel, exploration using `Make_seq` and `Make_iter` explore widely differently. Specifically, the `Seq`-based monad explores a lot less branches

giltho:
Found this while debugging the other issue

Resolves #8


N1ark (approved):
Oh!

PR #19, opened 2025-04-11
Fix in-place solver to properly handle reset

giltho:


N1ark (approved):
🎉

giltho:
@N1ark Can you check that the updated test outputs make sense?

N1ark:
Thanks! Seems a bit odd the final PCs are a bit smaller, is it because the fix pops more of the PC and the previous version was wrong?

giltho:
I think so! Supposedly, when you were running the second time, not all of the first run had been flushed. Now, everything is flushed! It also means that variable counters are getting properly reset so variable names have lower integers

PR #22, opened 2025-04-11
Perform substitution correctly for n-ary operators

giltho:


N1ark (approved):
oh never realised !

giltho:
That’s because you don’t use substitutions in wpst

PR #23, opened 2025-04-11
Multi-file analysis for soteria-c

giltho:
I have extended bfa-c to support multi-file analysis.
Right now, it simply comes in the form of a command that looks like
```
bfa-c exec-main file1.c ... fileN.c -I include_folder_1 -I include_folder_2
```

The most important file of this PR is really `ail_linking.ml`  which follows [`core_linking.lem`](https://github.com/rems-project/cerberus/blob/df9e0a48d78bf0d6ffd98cf42555ad76fbc6a1e6/frontend/model/core_linking.lem) as best as I could.
I added a new type `linked_program` type which is a normal `AilSyntax.program` with, in addition, captures a `symmap`. The `symmap` is a union-find which maps each symbol to its real symbol declaration. In the interpreter, when resolving a symbol, I simply follow the union find until I reach the representative and call that.

Outstanding questions are:
- Should I also erase unused definitions in `sigma.definitions`?
- I'm not quite sure what the topo sort is doing, but I did it anyway. Is there anything I could have done wrong?


bfa_c/lib/ail_linking.ml
```diff
@@ -0,0 +1,299 @@
...
+      Sym_set.diff res exclude
+  | AilSpar stmtl -> List.fold_left free_syms_stmt acc stmtl
+
+let free_syms_object_definition acc expr = free_syms_expr acc expr
+
+let merge_globs (globs_1 : 'a sigma_object_definition list)
+    (globs_2 : 'a sigma_object_definition list)
+    (redundant_globs : (Symbol.sym * Symbol.sym) list) =
+  let catch f = try Ok (f ()) with LinkError s -> Error s in
+  catch @@ fun () ->
```
N1ark, line 156: nice trick!

bfa_c/lib/interp.ml
```diff
@@ -43,8 +43,8 @@ module Make (Heap : Heap_intf.S) = struct
         Csymex.return x
     | _ -> Fmt.kstr Csymex.not_impl "Not a pointer: %a" Typed.ppa x
 
-  type 'err fun_exec =
-    prog:sigma ->
+  type ('a, 'err) fun_exec =
+    prog:linked_program ->
     args:T.cval Typed.t list ->
     state:state ->
     (T.cval Typed.t * state, 'err, Heap.serialized list) Result.t
```
N1ark, line 50: Why did you add the `'a`?

bfa_c/lib/ail_linking.ml
```diff
@@ -0,0 +1,299 @@
...
+            (free_syms_object_definition (Sym_set.of_list acc) glob) ))
+      globs
+  in
+  (* init_set is the list of symbol with no dependencies *)
+  (* We order the symbols in topological order! *)
+  let ordered_syms =
+    match Tsort.sort dep_map with
+    | ErrorCycle _ -> raise (LinkError "cycle in global dependencies!")
+    | Sorted l -> l
+  in
```
N1ark, line 179: Yeah I'm not sure how needed the sort is but it catches cycles which (I think) you want to catch so it's nice! Shouldn't hurt

N1ark (approved):
Exciting, congrats! Small things but merge whenever-- about erasing unused definitions in `sigma.definitions`, I have no idea -- I think you can and it wouldn't hurt, but I think leaving them should be ok either way. From what I understand they'd be unreachable anyways, so removing them could save some perf if you're doing operations on the list.

bfa_c/lib/interp.ml
giltho: Ah, at some point I had `'a linked_program` with two different kinds of resolved programs, and then I gave up on that because I cluttered everything.

Good catch!

PR #26, opened 2025-04-15
Dynamic function call

giltho:
Requires merging #25


bfa_c/lib/driver.ml
```diff
@@ -173,6 +176,11 @@ let resolve_entry_point (linked : Ail_tys.linked_program) =
   |> Bfa_std.Utils.Result_ex.of_opt
        ~err:(`ParsingError "Entry point not found", Call_trace.empty)
 
+let with_function_context prog f =
+  let open Effect.Deep in
+  let fctx = Fun_ctx.of_linked_program prog in
+  try f () with effect Interp.Get_fun_ctx, k -> continue k fctx
+
```
N1ark, line 183: I don't know much about effects, could you briefly explain how this works?

bfa_c/lib/fun_ctx.ml
```diff
@@ -0,0 +1,50 @@
...
+          In any case, we will fix this when we have native support for disjoint addresses.
+*)
+type t = { counter : Z.t; bmap : Bidirectional_map.t }
+
+let empty = { counter = Z.zero; bmap = Bidirectional_map.empty }
+
+let declare_fn sym t =
+  let counter = t.counter in
+  let bmap = Bidirectional_map.add sym counter t.bmap in
+  { counter = Z.add counter Z.one; bmap }
```
N1ark, line 30:
```suggestion
  { counter = Z.succ counter; bmap }
```

N1ark (approved):
Nice! Not ssure how the effect works but overall seems good

bfa_c/lib/fun_ctx.ml
giltho: Ah, I tried `Z.incr` but it didn't work hahahaha

N1ark: I agreee `incr` is more intuitive haha

bfa_c/lib/driver.ml
giltho: Yes, effects are just "resumable exceptions"

So when you do like
```
exception Some_except

let _ = try f () with Some_except -> (* HERE *)
```
if `f ()` raises an exception, it jumps "HERE" and lets you handle the exception

An effect is exactly the same, but it also gives you a continuation (here, `k`), which allows you to resume the execution after handling the effect.

So, to give you a bit more intuition, exceptions can strictly be re-implemented using effects, by just never using the continuation it gives you

Here, I am using effects to fetch the context. I would obtain the exact same result by just putting the function context in a global reference. The main difference is: if we one day want to parallelise execution, all parallel branches would modify the same global reference and we'd get inconsistent results. On the other hand, the effect here is handled strictly in this execution, and is therefore thread-safe.

N1ark: ahhhhh ok yes got it !! wow that's neat yes i'll try doing this with `crate` to see if it impacts performance

giltho: Go on holiday or I block you from the github org

PR #29, opened 2025-04-16
Leak checks!

N1ark:
Add a flag `--leaks` to look for memory leaks

### DIFF
--- a/bfa_rust/scripts/miri.sh
+++ b/bfa_rust/scripts/miri.sh
@@ -109,11 +109,16 @@ for test in $TESTS; do
     # expect failure if "fail" or "panic" in test_rel_name
     expect_failure=$(grep -c -e "fail" -e "panic" <<< "$test_rel_name")
 
+    extra_flags=""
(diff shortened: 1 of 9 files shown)


bfa_rust/scripts/miri.sh
```diff
@@ -109,11 +109,16 @@ for test in $TESTS; do
     # expect failure if "fail" or "panic" in test_rel_name
     expect_failure=$(grep -c -e "fail" -e "panic" <<< "$test_rel_name")
 
+    extra_flags=""
+    if [[ $(grep -c "\-Zmiri\-ignore\-leaks" "$test") -eq 0 ]]; then
+        extra_flags="--leaks"
+    fi
+
```
giltho, line 116: Can we mirror Miri's own behaviour and have a `--ignore-leaks` flag instead of a `--leaks` flag?

N1ark: Yes sure! I just made this choice because e.g. Kani doesn't check for leaks, so the "default" behaviour for them is not checking, but yes mirroring Miri works too

giltho (approved):

PR #31, opened 2025-04-17
Cache Charon Binaries in CI

giltho:


giltho:
The last CI run is failing because running tests will actually.. install the correct toolchain 🤦 
image

What needs to be done is:
- Cache also the toolchain file
- When installing the rust toolchain, already select the one from the toolchain file given by charon, if there was no cache hit, we need to clone charon before we install the toolchain

PR #38, opened 2025-04-25
More rust

N1ark:
- Try splitting the montrosity that is `std_funs` into: `builtins/std`, `builtins/kani`, and `builtins/eval` that handles the identifier to function resolving.
- Add binary operators, via bit vectors
- Add some std functions
- Fix comparison of pointers with metadata
- Add `as_zst` to get a type as its possible ZST value -- this is used when attempting to load a ZST, to allow reading with invalid pointers
- In `eval_place`, parse `Base x` values as a null pointer with offset `x`
- add `kani::Invariant`


soteria-rust/lib/interp.ml
```diff
@@ -482,13 +542,50 @@ module Make (Heap : Heap_intf.S) = struct
...
+                        args = Some max;
+                      } ) ->
+                    let max = int_of_string max in
+                    let* v = cast_checked ~ty:Typed.t_int v in
+                    if%sat v <=@ Typed.int max then Result.ok ((), state)
+                    else
+                      Heap.error (`StdErr "rustc_layout_scalar_valid_range_end")
+                        state
+                | _ -> Result.ok ((), state))
+        in
```
N1ark, line 588: This whole thing is to pass one Miri test, and I don't particularly like how it's implemented, so depending on your opinion I might remove it or implement it another way -- Im just unsure how to do it more neatly, and I don't think many attributes need to be checked aside from these two...

soteria-rust/lib/encoder.ml
```diff
@@ -63,6 +64,7 @@ let rec rust_to_cvals ?(offset = 0s) (value : 'ptr rust_val) (ty : Types.ty) :
   | _, TAdt (TBuiltin TBox, _) | _, TRawPtr _ | _, TRef _ -> illegal_pair ()
   (* Tuples *)
   | Tuple vs, TAdt (TTuple, { types; _ }) -> chain_cvals (layout_of ty) vs types
+  | Tuple [], TNever -> []
```
giltho, line 67: Hmmm, where do you have a value of type `TNever`?
`TNever` should never be constructed

soteria-rust/lib/interp.ml
```diff
@@ -482,13 +535,50 @@ module Make (Heap : Heap_intf.S) = struct
...
     (* Struct aggregate *)
-    | Aggregate (AggregatedAdt (TAdtId _, None, None, _), operands) ->
-        let++ values, state = eval_operand_list ~crate ~store state operands in
+    | Aggregate (AggregatedAdt (TAdtId t_id, None, None, _), operands) ->
+        let type_decl =
+          Types.TypeDeclId.Map.find t_id UllbcAst.(crate.type_decls)
+        in
+        let** values, state = eval_operand_list ~crate ~store state operands in
+        let++ (), state =
+          if 0 <> List.compare_length_with values 1 then Result.ok ((), state)
```
giltho, line 549:
```ocaml
match values with
| [ v ] -> 
| _ -> Result.ok ((), state)
```
?

giltho (approved):

soteria-rust/lib/interp.ml
giltho: I actually find it quite nice that you support this.
Can you maybe just factor it out into another function?

soteria-rust/lib/rustsymex.ml
```diff
@@ -1,7 +1,7 @@
 module SYMEX =
   Soteria_symex.Symex.Make_iter
     (struct
-      let fuel : Soteria_symex.Fuel_gauge.t = { steps = 500; branching = 10 }
+      let fuel : Soteria_symex.Fuel_gauge.t = { steps = 300; branching = 4 }
```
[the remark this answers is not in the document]
N1ark: yeah....

soteria-rust/lib/builtins/kani.ml
```diff
@@ -0,0 +1,35 @@
...
+  type nonrec rust_val = Sptr.t rust_val
+
+  let assert_ ~crate:_ ~(args : rust_val list) ~state =
+    let open Typed.Infix in
+    let* to_assert =
+      match args with
+      | [ Base t; _msg ] -> cast_checked t ~ty:Typed.t_int
+      | _ -> not_impl "to_assert with non-one arguments"
+    in
+    if%sat to_assert ==@ 0s then Heap.error `FailedAssert state
```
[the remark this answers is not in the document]
N1ark: Done! No actually, quite straightforward

soteria-rust/lib/encoder.ml
N1ark: hmm not sure actually ! fixed

soteria-rust/lib/interp.ml
N1ark: ..... yes thats much more sensible

PR #40, opened 2025-04-29
Log filters

N1ark:


soteria/lib/logging/assets/soteria_logs.css
```diff
@@ -1,84 +1,100 @@
 body {
```
giltho, line 1: We need some file to settle CSS style otherwise this is going to come up in every PRs

giltho (approved):

soteria/lib/logging/assets/soteria_logs.css
N1ark: yes :) next time changes are needed we can add a .prettierrc

PR #41, opened 2025-04-29
Rust Bit Vectors

N1ark:
Not super satisfied yet, but i think this is ok for a first increment; in particular it would be interesting to see if i can remove `IntOfBv`, and simply keep the value as a BV and decay everything around it as needed.

For now, this at least improves performance of some binary operator tests, as well as any code that uses floating point functions like `is_normal`, `classify`, etc ([list](https://doc.rust-lang.org/nightly/std/primitive.f64.html#method.is_nan))

Also, properly detect `#[rustc_intrinsic]` and vanish if unsupported


soteria-rust/lib/builtins/std.ml
```diff
@@ -693,20 +693,15 @@ module M (Heap : Heap_intf.S) = struct
...
-            in
             (* convert to unsigned *)
             let v =
               if Layout.is_signed ty then
                 let maxv = Z.shift_left Z.one bits in
                 if Z.(v < zero) then Z.(((v mod maxv) + maxv) mod maxv) else v
               else v
             in
-            Result.ok (Base (aux 0 v), state)
+            let v = Typed.int @@ Z.popcount v in
```
giltho, line 703: 😆

giltho (approved):
lgtm

soteria-rust/lib/builtins/std.ml
N1ark: when i was looking at the signatures in Z and realised ... felt silly

PR #43, opened 2025-05-01
Even more Rust: Charon, DSTs, splitting, cleaning

N1ark:
- Bump Charon
- Removed a bunch of builtin functions that had no reason to be builtin
- Add `intrinsics::variant_count`
- Add Kani's STD library, that overrides `assert!`, `panic!`, etc.
- Add `--kani` flag to add the Kani library, add `--miri` flag to add the (to be implemented) Miri library
- Properly handle casting from pointers to integers, by adding `Sptr.decay`
- Similarly, handle casting from integer to pointers by making them a null pointer with the integer as an offset -- using it is not valid of course, but creating it is (for raw ptrs)
- Add splitting nodes (for integers) -- this required some changes to TreeBlock.{split,extract}. We assume a little-endian system.
- Simplify binary operations of the form `(IntOfBv bv == n)` into `(bv == BvOfInt n)` when `n` is concrete
- Simplify binary operations of the form `(X & #x...N) = #x...M` to `(X & #xN) = #xM`, ignoring padding when using a mask
- Add `Option` to Soteria's Stdlib
- Handle dynamically sized types better; the fix is simple but dodgy, so I'm open to comments/improvements:
  I used to error when computing the layout of a DST; however that meant e.g. computing the layout of `{ f1: T1, ..., fn: [TN] }` was not possible, despite being allowed.
  Instead, I now consider that `[T]` has a layout of size 0 and align `align_of<T>` -- this means the layout of the above struct can be computed, and reading it works too without further changes.
  However this *is* a little dodgy, possibly.


soteria/lib/soteria-std/option.ml
```diff
@@ -0,0 +1,10 @@
+include Stdlib.Option
+
+let bind2 f x y = match (x, y) with Some x, Some y -> f x y | _ -> None
+let map2 f x y = match (x, y) with Some x, Some y -> Some (f x y) | _ -> None
```
giltho, line 4: This feels like you'd be better off defining `and*`?

```ocaml
  let bind2 f x y = let* x and* y in f x y
  
  let map2 f x y = let+ x and* y in f x y
```

giltho (approved):
A lot of this is a bit difficult to review further than just OCaml comments, but the amount of features you keep adding is impressive!

soteria/lib/soteria-std/option.ml
N1ark: oh I forget `and*` is a thing!
I tried this, but it would mean defining `and*` in Syntaxes so then I cant use it in `map2`/`bind2`, and using it directly in the relevant code is quite hard to parse for me at least:
```ocaml
let msb =
  let+ v1 = msb_of v1 and* v2 = msb_of v2 in
  max v1 v2
in ...
```

N1ark: I'll leave it like this for now, but open to change it in the future

PR #45, opened 2025-05-16
Add Plugin system (better Miri support)

N1ark:
- Add a `rusteria` library, that comes with all executions and gives access to basic symbolic execution constructs: `nondet`, `assume`, `assert` and `panic` 
  `panic` just panics -- i might remove it later if it causes issues with unwind, but it's nicer for errors for now
- Move Kani's "std" (overriding some macros like `assert!`, remove printing stuff) to `rusteria`, so  non-Kani tests also enjoy from a simpler MIR.
- Add a "plugin" system (better name ideas?), that adds steps to the compilation of tests dynamically. Plugins also allow custom detection of entry points (e.g. for Kani with `#[kani::proof]`, and Miri with `miri_start`).
- Bump Charon to one of my branches to avoid some parsing errors (will be fixed in https://github.com/AeneasVerif/charon/pull/685)
- Add support for a few Miri builtins ! Not all for now but enough to run some interesting tree borrow tests.

One thing I, for now, can't seem to get to work is allow compiled code to access `core` -- a bunch of Miri tests try to use it and `rustc` tells me the crate isnt found and suggests using `std` instead. So for now to run the Miri tests I created a branch that uses our `miristd`, and that replaces all `core`s with `std`: https://github.com/N1ark/miri/tree/rusteria. It works, except for the handful of `no_std` tests.


soteria-rust/lib/plugin.ml
```diff
@@ -0,0 +1,164 @@
...
+      verbosity
+  in
+  if res <> 0 && res <> 255 then
+    let msg = Fmt.str "Couldn't compile lib at %s: error %d" path res in
+    raise (PluginError msg)
+
+type plugin = {
+  mk_cmd : unit -> Cmd.charon_cmd;
+  get_entry_point : fun_decl -> entry_point option;
+}
```
N1ark, line 46: fyi i tried doing this as a module type, and then generating modules at runtime, but the syntax was really not nice so i avoided it, felt like over-engineering anyways :p

soteria-rust/lib/builtins/eval.ml
```diff
@@ -51,13 +55,22 @@ module M (Heap : Heap_intf.S) = struct
...
     [
-      (* Kani *)
-      ("kani::assert", KaniAssert);
-      ("kani::assume", KaniAssume);
-      ("kani::nondet", KaniNondet);
+      (* Rusteria builtins *)
+      ("rusteria::assert", RusteriaAssert);
+      ("rusteria::assume", RusteriaAssume);
+      ("rusteria::nondet", RusteriaNondet);
+      ("rusteria::panic", RusteriaPanic);
```
giltho, line 62: Is there a reason why someone would want to use Rusteria but not Kani's interface?

soteria-rust/lib/builtins/eval.ml
```diff
@@ -51,13 +55,22 @@ module M (Heap : Heap_intf.S) = struct
...
+      ("kani::assert", RusteriaAssert);
+      ("kani::panic", RusteriaPanic);
+      (* Miri builtins *)
+      ("miristd::miri_get_alloc_id", MiriAllocId);
+      ("miristd::miri_pointer_name", Nop);
+      ("miristd::miri_print_borrow_state", Nop);
       (* Core *)
       (* FIXME: get rid of these, as Charon improves *)
       ("alloc::boxed::{alloc::boxed::Box}::into_raw", BoxIntoRaw);
+      ("alloc::boxed::{@T}::from_raw", BoxIntoRaw);
```
giltho, line 73: Typo?

giltho (approved):

soteria-rust/lib/builtins/eval.ml
N1ark: no no it's just that BoxIntoRaw just returns the input ptr, and BoxFromRaw technically does the exact same, so i didn't want to redo it :p

soteria-rust/lib/builtins/eval.ml
N1ark: the main reason i did the separation is bc when i override std functions i need an assert fn as well as a panic, and id rather not import all of kani's lib when testing miri :p 
but yes i think its nice to expose a symbolic lib that is simpler than kani's, we could later add more specific stuff to it if we get new ideas

PR #48, opened 2025-05-20
add collections-c test in CI, upgrade aggregate_cc.fish script

giltho:
Depends on #47, but I wanna run CI now


giltho:
@N1ark This one is ready for review now :)

PR #49, opened 2025-05-21
Fix bug, add interface to execute only one function

giltho:


giltho:
@N1ark last one 😬 (Not necessarily today)

N1ark (approved):

PR #50, opened 2025-05-23
Extract c_values into a re-usable module

giltho:


N1ark (approved):
Nice! Excited to update the Rust side too, to get rid of this duplicated mess

PR #51, opened 2025-05-23
Support `Cell`, better tree borrows, fix unions

N1ark:
Sorry i really thought this would be a smaller one :'/ a small problem we have is that adding an operation to treeblock incurs so much added code, since you need to wrap the `Tree` function at the `Tree_block` level, then wrap that at the `Heap` level, then add it to the intf.... but i guess there's no choice :p would be cool to get a simpler state model in the future and see how well it does and how much easier it is to maintain, not that this is hard to maintain but its verbose

- Support `UnsafeCell` for tree borrows:
  - Any type containing an `UnsafeCell` is considered interiorly mutable (in its entirety)
  - Shared borrows of `&UnsafeCell<T>` inherit the parent's tag
  - Mutable borrows of `&UnsafeCell<T>` start in ReservedIM
- Add some missing tree borrows features:
  - All references passed as arguments receive a read; this was already the case for `&T`, but not e.g. for `Option<&T>`, or other nested cases
  - Similarly, all references returned also receive a read to ensure they're still valid
  - Protectors are only toggled on the area of the references type at function entry, but using that reference on any area outside its bounds also toggles the protector for it; this change means that the tree borrow state in tree blocks doesn't only store the state of each tag, but also whether it has a protector
  - Added `protect`, `unprotect` and `tb_access` (ghost read, without validating the values) operations to Heap/TreeBlock to enable all this
- Made the parser in `encode.ml` a proper-ish monad. Made it handle unions semi-properly (/better): will sort the fields of the union by size, and will try reading all fields until one of them matches. 
  We sort the fields in decreasing size, to make sure in a case like `union Foo<T> { a: (), b: T }` the unit doesn't always get parsed (e.g. for `MaybeUninit`)


giltho:
Yeah I agree re treeblocks, it's verbose. There's probably a way to do better, but at the moment it's honestly not too hard even though it's verbose

soteria-rust/lib/heap.ml
```diff
@@ -128,66 +128,75 @@ let check_ptr_align (ptr : Sptr.t) ty =
...
 
 let uninit (ptr, _) ty st =
   let@ () = with_error_loc_as_call_trace () in
   let@ () = with_loc_err () in
   log "uninit" ptr st;
   let* size = Layout.size_of_s ty in
-  with_ptr ptr st (fun ~ofs block ->
-      let@ block, _ = with_tbs block in
-      Tree_block.uninit_range ofs size block)
+  let@ ofs, block = with_ptr ptr st in
```
giltho, line 148: yep

soteria-rust/lib/interp.ml
```diff
@@ -861,8 +878,8 @@ module Make (Heap : Heap_intf.S) = struct
     in
     let@ () = with_loc ~loc in
     L.info (fun m ->
-        m "Calling %s with [@[%a@]]" (name_str crate name)
-          Fmt.(list ~sep:comma pp_rust_val)
+        m "Calling %s with [%a]" (name_str crate name)
+          Fmt.(list ~sep:(any ", ") pp_rust_val)
```
giltho, line 882:
```suggestion
        m "Calling %s with %a" (name_str crate name)
          Fmt.(hbox @@ brackets @@ list ~sep:comma pp_rust_val)
```

giltho: Will force to be on a single line using proper fmt combinators

soteria-rust/lib/layout.ml
```diff
@@ -104,6 +104,11 @@ module Session = struct
     | Struct fields -> fields
     | _ -> assert false
 
+  let as_union adt_id =
+    match (get_adt adt_id).kind with
+    | Union fields -> fields
+    | _ -> assert false
```
giltho, line 110:
```suggestion
    | _ -> failwith "unreachable: as_union"
```

soteria-rust/lib/tree_block.ml
```diff
@@ -766,6 +807,36 @@ let zero_range ofs size t =
...
+
+let tb_access ofs size tag tb t =
+  match t with
+  | None -> miss_no_fix ~msg:"protect on none" ()
+  | Some t ->
+      let++ (), tree =
+        let@ () = with_bound_check t (ofs +@ size) in
+        Tree.tb_access ofs size tag tb t.root
+      in
+      ((), to_opt tree)
```
giltho, line 838: - You forgot to fix the message in each copy-pasted line
- None-check + bound check can be factored out into a single function

giltho (approved):
minor changes

soteria-rust/lib/layout.ml
N1ark: I'll update this in the branche with the `Crate` module, I don't want merge conflicts :p

PR #58, opened 2025-05-27
More SValues !

N1ark:
- Type operators for floats: `==`, `<`, `<=`, `+`, `-`, `*`, `/`, `rem`, `abs`
- Add floating-point class unop: `FIs {NaN,Inf,Normal,Subnormal,Zero}` -- Rust has functions that do this with bitwise operations but that's 1. super slow and 2. broken, for some reason
- Add bitvector `+` and `-`, which allows usually deeper conversions in `bv_of_int`
- Make most functions in `SValue` recursive and chain them with `and`
  This is just because they often end up calling each other and having to move them to make things work was getting annoying so yeah...

### DIFF
--- a/soteria-rust/lib/interp.ml
+++ b/soteria-rust/lib/interp.ml
@@ -427,21 +427,34 @@ module Make (Heap : Heap_intf.S) = struct
         match (v1, v2) with
         | Base v1, Base v2 -> (
             match op with
-            | Ge | Gt | Lt | Le ->
-                let op =
-                  match op with
-                  | Ge -> Typed.geq
-                  | Gt -> Typed.gt
-                  | Lt -> Typed.lt
-                  | Le -> Typed.leq
-                  | _ -> assert false
-                in
+            | Ge | Gt | Lt | Le -> (
                 let* v1, v2, ty = cast_checked2 v1 v2 in
-                if Typed.equal_ty ty Typed.t_ptr then
-                  Heap.error `UBPointerComparison state
-                else
-                  let v = op v1 v2 |> Typed.int_of_bool in
-                  Result.ok (Base v, state)
+                match Typed.untype_type ty with
+                | Svalue.TInt ->
+                    let op =
+                      match op with
+                      | Ge -> Typed.geq
+                      | Gt -> Typed.gt
+                      | Lt -> Typed.lt
+                      | Le -> Typed.leq
+                      | _ -> assert false
+                    in
+                    let v = op v1 v2 |> Typed.int_of_bool in
+                    Result.ok (Base v, state)
+                | TFloat _ ->
+                    let op =
+                      match op with
+                      | Ge -> Typed.geq_f
+                      | Gt -> Typed.gt_f
+                      | Lt -> Typed.lt_f
+                      | Le -> Typed.leq_f
+                      | _ -> assert false
+                    in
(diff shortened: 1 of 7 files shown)


soteria/lib/c_values/svalue.ml
```diff
@@ -410,38 +468,38 @@ let rec sem_eq v1 v2 =
         sem_eq bv (BitVec n <| t_bv size)
     | _ -> mk_commut_binop Eq v1 v2 <| TBool
 
-let sem_eq_untyped v1 v2 =
+and sem_eq_untyped v1 v2 =
   if equal_ty v1.node.ty v2.node.ty then sem_eq v1 v2 else v_false
 
-let or_ v1 v2 =
+and or_ v1 v2 =
```
giltho, line 474: Wait why is everything co-recursive, you're not calling anything else in `or_`?

soteria/lib/c_values/svalue.ml
```diff
@@ -410,38 +468,38 @@ let rec sem_eq v1 v2 =
...
   match (v1.node.kind, v2.node.kind) with
   | Bool b1, Bool b2 -> bool (b1 || b2)
```
giltho, line 476: Redundant with the next cases. If you have a boolean on either side, then you have a boolean on both sides.
Also, the other cases avoid reallocating a new boolean, so they're cheaper.

soteria-rust/lib/interp.ml
```diff
@@ -421,21 +421,34 @@ module Make (Heap : Heap_intf.S) = struct
...
-                  Result.ok (Base v, state)
+                match Typed.untype_type ty with
+                | Svalue.TInt ->
+                    let op =
+                      match op with
+                      | Ge -> Typed.geq
+                      | Gt -> Typed.gt
+                      | Lt -> Typed.lt
+                      | Le -> Typed.leq
+                      | _ -> assert false
```
giltho, line 434: In general, I like to avoid assert false and use failwith "...." so that if something goes wrong because it is not actually unreachable, I have something more precise than assert false.

soteria-rust/lib/interp.ml
```diff
@@ -421,21 +421,34 @@ module Make (Heap : Heap_intf.S) = struct
...
-                  | _ -> assert false
-                in
+            | Ge | Gt | Lt | Le -> (
                 let* v1, v2, ty = cast_checked2 v1 v2 in
-                if Typed.equal_ty ty Typed.t_ptr then
-                  Heap.error `UBPointerComparison state
-                else
-                  let v = op v1 v2 |> Typed.int_of_bool in
-                  Result.ok (Base v, state)
+                match Typed.untype_type ty with
```
giltho, line 426: Somehow feels like you should have constructors that do the correct typed thing depending on the lhs type

giltho (approved):
Why the massive co-recursive definitions?

soteria/lib/c_values/svalue.ml
N1ark: true! fixed it in and_ too

soteria-rust/lib/interp.ml
N1ark: yeah... i only do it here and i have something to handle the same for + - * / in Core, so i didnt bother too much, it doesnt happen often at all and the functions are typed so it's relatively safe

PR #56, opened 2025-05-27
Merge plugins, add `Crate`

N1ark:
- Merge all plugins into one dune site
- Add a `Crate` module to effectfully pass the current crate (beautiful)

PR #59, opened 2025-05-27
More and more and more Kani, but cleaned up : ) 

N1ark:
Same as before, but now depends on: #56 #57 #58

### DIFF
--- a/soteria-rust.opam
+++ b/soteria-rust.opam
@@ -52,7 +52,7 @@ build: [
 dev-repo: "git+https://github.com/soteria-tools/soteria.git"
 pin-depends: [
   ["simple_smt.~dev" "git+https://github.com/NatKarmios/simple-smt-ocaml#9db7ef0bb61a76010de255af3cbe4066dd8dd622"]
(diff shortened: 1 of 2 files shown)


soteria-rust/lib/core.ml
```diff
@@ -36,10 +36,11 @@ module M (Heap : Heap_intf.S) = struct
...
       wrapping behaviour.) *)
-  let safe_binop (bop : Expressions.binop) (l : [< T.cval ] Typed.t)
-      (r : [< T.cval ] Typed.t) st : ([> T.cval ] Typed.t, 'e, 'm) Result.t =
-    let* l, r, ty = cast_checked2 l r in
-    match untype_type ty with
+  let safe_binop (bop : Expressions.binop) (ty : Types.literal_type)
+      (l : [< T.cval ] Typed.t) (r : [< T.cval ] Typed.t) st :
+      ([> T.cval ] Typed.t, 'e, 'm) Result.t =
+    let* l, r, ty_ = cast_checked2 l r in
+    match untype_type ty_ with
```
giltho, line 43: At some point we really need to add those behind abstractions or syntax

PR #60, opened 2025-05-27
Propose a non-incremental Z3solver

giltho:
Signed-off-by: Sacha-Élie Ayoun <sachaayoun@gmail.com>


soteria/lib/c_values/z3solver_noincr.ml
```diff
@@ -0,0 +1,601 @@
...
+let memoz table f v =
+  match Hashtbl.Hint.find_opt table v.Hashcons.tag with
+  | Some k -> k
+  | None ->
+      let k = f v in
+      Hashtbl.Hint.add table v.Hashcons.tag k;
+      k
+
+let rec encode_value (v : Svalue.t) =
+  match v.node.kind with
```
N1ark, line 420: could you maybe port the SValue->sexp encoding to a separate file shared among both?

N1ark (changes requested):
Maybe just less code duplication, but so so so exciting !! ill try running a benchmark w my kani to see how it runs

soteria/lib/c_values/c_solver.ml
```diff
@@ -0,0 +1,318 @@
...
+        Dynarray.find_map
+          (fun { value; _ } ->
+            if Typed.equal value v then Some true
+            else if Typed.equal value neg_v then Some false
+            else None)
+          d)
+      t
+
+  let iter (t : t) f =
+    Dynarray.iter (fun t -> Dynarray.iter (fun { value; _ } -> f value) t) t
```
N1ark, line 78:
```suggestion
    Dynarray.iter (Dynarray.iter (fun { value; _ } -> f value)) t
```

soteria/lib/c_values/c_solver.ml
```diff
@@ -0,0 +1,318 @@
...
+
+  let to_value_list (t : t) = Iter.to_rev_list (iter t)
+
+  let mem (t : t) v =
+    Dynarray.exists
+      (fun d -> Dynarray.exists (fun { value; _ } -> Typed.equal v value) d)
+      t
```
N1ark, line 85:
```suggestion
    Dynarray.exists
      (Dynarray.exists (fun { value; _ } -> Typed.equal v value))
      t
```

soteria/lib/c_values/c_solver.ml
```diff
@@ -0,0 +1,318 @@
...
+    let changed = ref false in
+    let var_set = Var.Hashset.with_capacity 8 in
+    let vars value = Value.iter_vars value |> Iter.map fst in
+    let to_encode = Dynarray.create () in
+    let add_vars_raw vars = Var.Hashset.add_iter var_set vars in
+    let add_vars vars =
+      vars @@ fun v ->
+      if not (Var.Hashset.mem var_set v) then (
+        changed := true;
+        Var.Hashset.add var_set v)
```
N1ark, line 120: this is ok, but is there not a way to add and have the hashset say if the element was actually added? feels silly to only add if `not mem`, when that's the point of a hashset

soteria/lib/c_values/c_solver_incr.ml
```diff
@@ -0,0 +1,175 @@
...
+    let neg_v = Typed.not v in
+    Dynarray.find_map
+      (fun d ->
+        Dynarray.find_map
+          (fun value ->
+            if Typed.equal value v then Some true
+            else if Typed.equal value neg_v then Some false
+            else None)
+          d)
+      t
```
N1ark, line 59:
```suggestion
    Dynarray.find_map
      (Dynarray.find_map (fun value ->
           if Typed.equal value v then Some true
           else if Typed.equal value neg_v then Some false
           else None))
      t
```

N1ark (approved):
nits, exciting !! will benchmark asap

soteria/lib/c_values/c_solver.ml
N1ark: check the implementation of Hashset.len, if it's constant time you can use that to avoid a second traversal, just check if size increased

N1ark: ie ```ocaml
let before = Hashset.cardinal var_set in
let () = HashSet.add var_set v in
if before <> Hashset.cardinal var_set then changed := true;
```

giltho: Ah I did that in parallel didn't see that comment

PR #62, opened 2025-05-29
Add `eval` and `eval-diff` commands to script

N1ark:
Add an `eval` command to `test.py` to repeatedly run non-erroring tests, to measure performance changes. Iteration count can be controlled with the `-i` flag. Produces a `.csv` flag with the average execution time for each test.

image

Add an `eval-diff` command, that receives two files and prints the performance difference between the two runs.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -230,6 +230,7 @@ __pycache__/
 
 ### Produced by Soteria ###
 json_dump/
+soteria-rust/scripts/*.csv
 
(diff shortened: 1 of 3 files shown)


giltho (approved):

giltho:
Really nice! (Obviously, for script PRs, I don't read the code in as much details as I do for OCaml core code)

N1ark:
> Really nice! (Obviously, for script PRs, I don't read the code in as much details as I do for OCaml core code)

yeah of course, i can imagine :)

PR #65, opened 2025-06-02
Add dynamic function call

N1ark:
Depends on #64 

- Support fonction pointers and dynamic function calls
- Add `std::intrinsics::catch_unwind`, which now works !
- Minor printing changes


soteria-rust/lib/interp.ml
```diff
@@ -143,9 +143,9 @@ module Make (Heap : Heap_intf.S) = struct
             let** (), state = Heap.store ptr str_ty char_arr state in
             let++ (), state = Heap.store_str_global str ptr state in
             (Ptr ptr, state))
+    | CFnPtr fn_ptr -> Result.ok (ConstFn fn_ptr, state)
     | CLiteral (VByteStr _) -> not_impl "TODO: resolve const ByteStr"
     | CTraitConst _ -> not_impl "TODO: resolve const TraitConst"
-    | CFnPtr _ -> not_impl "TODO: resolve const FnPtr"
```
giltho, line 148: image

giltho (approved):

soteria-rust/lib/interp.ml
N1ark: 🤣

PR #68, opened 2025-06-04
Styling for progress bar

giltho:


N1ark:
enjoy the conflicts

soteria-c/lib/my_progress.ml
```diff
@@ -1,22 +1,75 @@
-let bar ~msg ~total =
+let rec string_contains ~sub_str = function
+  | "" -> sub_str = "" (* the empty string contains itself *)
+  | s ->
+      String.starts_with ~prefix:sub_str s
+      || string_contains ~sub_str (String.sub s 1 (String.length s - 1))
```
N1ark, line 5: baffled this isn't in the stdlib

soteria-c/lib/my_progress.ml
```diff
@@ -1,22 +1,75 @@
...
+  let is_ansi = is_term "ansi" in
+
+  match (term, color_term) with
+  | _, Some "true"
+  | _, Some "truecolor"
+  | Some ("xterm-kitty" | "wezterm"), _
+  | Some "linux", _ ->
+      true
+  | Some _, _ when is_256color || is_color || is_ansi -> true
+  | _ -> false
```
N1ark, line 24: what is a PR review without some styling nitpicks
```suggestion
  match (term, color_term) with
  | _, Some ("true" | "truecolor")
  | Some ("xterm-kitty" | "wezterm" | "linux"), _ ->
      true
  | Some _, _ when is_256color || is_color || is_ansi -> true
  | _ -> false
```

soteria-c/lib/my_progress.ml
giltho: It's in `Str`, which is distributed with the compiler but must be imported separately. Because Str uses a global state and isn't thread safe 💀

PR #69, opened 2025-06-05
Beautiful error messages ✨✨✨✨

giltho:


soteria-c/test/cram/simple_bi.t/run.t
```diff
@@ -27,22 +28,67 @@
...
+   16 │  int test_uninit()
+   17 │  {
+   18 │    int *x = (int *)malloc(sizeof(int));
+   19 │    if (!x)
+   20 │      return 1;
+   21 │    load(x);
+      │    ------- Called from here
+  error: NullDereference in test_np
+      ┌─ manifest.c:51:3
+   51 │    *x = 12;
```
N1ark, line 90: two nit picks:
1. Could we maybe add an empty line between diagnostics? bc here for the "`error:`" it's not really clear if it refers to the diagnostic above or below
2. For cases where there are a bunch of lines between the call site and the trigger mem location, does `grace` allow skipping some lines? e.g. here lines ~9-~17 we don't care about

giltho: 1. Yes, I wanted to do that and forget
2. I don't think so... I wanted but couldn't find it immediately in the API, but will look further just in case

PR #71, opened 2025-06-06
Add `grace` for Rust

N1ark:


soteria-rust/lib/dune
```diff
@@ -14,6 +14,9 @@
   iter
   zarith
   hashcons
+  grace
+  grace.ansi_renderer
+  spectrum
```
giltho, line 19: Is it really worth importing new library here? Fmt already has style formatting. Not as easy to write as spectrum but it might be worth trying to keep dependencies under control?

soteria-rust/test/cram/treeborrow.t/run.t
```diff
@@ -1,39 +1,58 @@
...
+    3 │ ╭  fn main() {
+    4 │ │      let mut root = 42;
+    5 │ │      let ptr = &mut root as *mut i32;
+    6 │ │      let (x, y) = unsafe { (&mut *ptr, &mut *ptr) };
+    7 │ │      *x = 13;
+    8 │ │      *y = 20; // UB: y is disabled
+      │ │      ^^^^^^^ Triggering memory operation
+    9 │ │      let val = *x;
+   10 │ │  }
+      │ ╰──' Call trace
```
giltho, line 24: Call trace captures the entire function?

giltho (approved):

soteria-rust/lib/dune
giltho:
```ocaml
let pp_string_red = Fmt.styled (`Fg `Red) Fmt.string
```
You don't have syntactic beauty of `@{<red> ...}` but same power

PR #77, opened 2025-06-09
Add `--ignore-aliasing`

N1ark:
Add `--ignore-aliasing`, which turns off aliasing checks.

It's done in a very lazy way: if disabled, borrows and protecting pointers do nothing, that's it, but it's sufficient; it means all pointers have the tag of the root pointer for that object, in state `Unique`, which cannot cause UB. 

Read/write access will be checked and all as usual, but since the tree will be a singleton and the state will only have one tag, the performance impact is minimal: for the SLL test, goes from ~25% of cycles to 2%. I can reduce it more in the future if it becomes an issue again, though ideally we should just have more performant tree borrows


demo/sll.rs
```diff
@@ -202,8 +202,8 @@ fn main() {
 
 #[kani::proof]
 #[kani::unwind(11)]
-#[rusteria::step_fuel(5000)]
-#[rusteria::branch_fuel(10)]
+#[cfg_attr(rusteria, rusteria::step_fuel(5000))]
+#[cfg_attr(rusteria, rusteria::branch_fuel(10))]
```
giltho, line 206: Nice!

soteria-rust/lib/heap.ml
```diff
@@ -361,7 +361,8 @@ let load_global g ({ globals; _ } as st) =
 let borrow (ptr, meta) (ty : Charon.Types.ty)
     (kind : Charon.Expressions.borrow_kind) st =
   (* &UnsafeCell<T> are treated as raw pointers, and reuse parent's tag! *)
-  if Layout.is_unsafe_cell ty then Result.ok ((ptr, meta), st)
+  if (kind = BShared && Layout.is_unsafe_cell ty) || Tree_borrow.is_disabled ()
```
giltho, line 364: Not sure how expensive Layout.is_unsafe_cell is, but `||` is short-circuiting in OCaml, so maybe the other way around would be quicker?

giltho (approved):

soteria-rust/lib/heap.ml
giltho: Just looked, it's definitely more expensive in some cases to have `is_unsafe_cell` since you're recursively going through the structure

PR #81, opened 2025-06-17
Interpreter Monad for Rust

N1ark:
more monads !! more monads !!!!

did the eval on 3 iterations of the kani suite; i get a +1.7%, so basically no difference :p


giltho:
Hmm, +1.7% is not "no difference", we need to be careful that we don't compound these slowdowns
But I think it's acceptable here

N1ark:
> Hmm, +1.7% is not "no difference", we need to be careful that we don't compound these slowdowns But I think it's acceptable here

sorry i didnt write this properly :') we get an improvement of 1.7%, so time is **-** 1.7% !!

soteria-rust/lib/builtins/miri.ml
```diff
@@ -7,7 +7,7 @@ module M (Heap : Heap_intf.S) = struct
 
   type nonrec rust_val = Sptr.t rust_val
 
-  let alloc_id ~args ~state =
```
giltho, line 10: Would it be worth defining InterpM outside Interp.ml so you can re-use it in miri/core/std?

soteria-rust/lib/builtins/std.ml
```diff
@@ -271,12 +275,12 @@ module M (Heap : Heap_intf.S) = struct
       let++ ptr' = Sptr.offset ~check ptr v |> Heap.lift_err state in
       (Ptr (ptr', meta), state)
 
-  let box_into_raw ~args ~state =
+  let box_into_raw ~args state =
     (* internally a box is exactly a pointer so nothing to do *)
     let box_ptr = List.hd args in
     Result.ok (box_ptr, state)
```
giltho, line 281: Cause like
```ocaml
let box_into_raw ~args = InterpM.ok (List.hd args)
```

giltho (approved):

soteria-rust/lib/builtins/std.ml
N1ark: ill try this in a separate pr/branch

PR #94, opened 2025-07-10
Log all our known failures

N1ark:
You were asking the reason why tests failed (success for fail, or fail for success -- ie not caused by an execution issue) -- here's the current list :) 
Includes all the Kani and Miri, aside from 6 TB issues that would take a bit longer to investigate so I didn't bother atm.
Fixed a couple of simple ones along the way


soteria-rust/scripts/common.py
```diff
@@ -116,31 +116,80 @@ def pptable(rows: list[list[tuple[str, Optional[str]]]]):
 
 KNOWN_ISSUES = {
     # Kani
+    "Arbitrary/arbitrary_structs.rs": "any_array isn't correctly implemented, due to mono issues",
     "ArithOperators/unsafe_add_fail.rs": "The main function takes a parameter? Kani crashes too",
     "ArithOperators/unsafe_mul_fail.rs": "The main function takes a parameter? Kani crashes too",
     "ArithOperators/unsafe_sub_fail.rs": "The main function takes a parameter? Kani crashes too",
+    "Drop/drop_after_mutating_refcell.rs": "We don't do drops properly",
+    "Drop/drop_concrete.rs": "We don't do drops properly",
+    "Drop/drop_enum_only_one_called.rs": "We don't do drops properly",
```
giltho, line 125: You wrote these by hand right?

N1ark: hahahah yes :p failure cause isn't rly easy to detect automatically i think because any number of thing could [not] go wrong

PR #98, opened 2025-08-05
Make fuel manipulation an effectful computation

giltho:
Here's a nice PR ;)


N1ark (approved):
nice!! looks good to me -- did you see any difference in performance?

PR #100, opened 2025-08-07
Derive Soteria-C Config's cmdliner term using a ppx

giltho:
This is PR #100 🎉 

Based on #99 

- Use ppx-subliner
- Make config retrieval an effect :)

### DIFF
--- a/soteria-c/lib/config.ml
+++ b/soteria-c/lib/config.ml
@@ -6,59 +6,64 @@ type t = {
         | [ x ] -> x
         | _ -> failwith "Multiple auto-include paths found"]
       [@docv "PATH"]
-      [@names [ "auto-include-path" ] [@env "SOTERIA_AUTO_INCLUDE_PATH"]]
+      [@names [ "auto-include-path" ]]
+      [@env "SOTERIA_AUTO_INCLUDE_PATH"]
       (** Path to the directory that contains the soteria-c.h *)
   dump_unsupported_file : string option;
       [@docv "FILE"]
-      [@names [ "dump-unsupported" ] [@env "SOTERIA_DUMP_UNSUPPORTED"]]
+      [@names [ "dump-unsupported" ]]
+      [@env "SOTERIA_DUMP_UNSUPPORTED"]
       (** Dump a json file with unsupported features and their number of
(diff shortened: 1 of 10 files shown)


soteria-c/lib/config.ml
```diff
@@ -1,24 +1,81 @@
...
+  havoc_undefined_funs : bool;
+      [@make.default false]
+      [@names
+        [ "havoc-undef"; "havoc-undefined-funs" ]
+        [@env "SOTERIA_HAVOC_UNDEFINED_FUNS"]]
+      (** Assume that all undefined functions can return any value. Warning:
+          this can lead to unsoundnesses in analyses. *)
 }
-[@@deriving make]
+[@@deriving make, subliner]
```
N1ark, line 71: beautiful

soteria-c/lib/config.ml
```diff
@@ -1,24 +1,81 @@
...
 
 let default = make ()
-let current : t ref = ref default
-let set (config : t) = current := config
+
+let current () =
+  try Effect.perform GetConfig with Effect.Unhandled GetConfig -> default
+
+let with_config ~(config : t) f =
+  try f () with effect GetConfig, k -> Effect.Deep.continue k config
```
N1ark, line 81: this is cool but also does it actually matter? as in, are there cases where we want several parallel configs? still good to have though :)

N1ark (commented):
lgtm! i dont think ill do the config-effect change on the rust side; and if I do i think by taste I won't have a `current ()` fn but instead just a bunch of functions for each config e.g. `Config.property ()` just bc the `(Config.current ()).property` looks a bit ugly; maybe a ppx? 👀

N1ark (changes requested):
pause !111

soteria-c/lib/config.ml
giltho: I think in general I'm uncomfortable with global mutable stuff. That's one of the things I really liked with Rust, and it's even easier to do with OCaml now. Given that the cost is none 🤷

N1ark: fair enough :)

PR #103, opened 2025-08-10
Charon mega-bump 3000

N1ark:
Still a chunky PR but more manageable; I'm not sure how to split this further, as it would be equivalent to opening 1 PR/commit kind of 
- Update Charon; globals are now places rather than rvalues!
- Remove `rust_val` & co. from `Charon_util.ml`, and make a `Rust_val.ml`
- Fixes to the Rust interpreter:
	- Fix `ShallowInitBox`
	- Fix `CopyNonOverlapping` copying tree borrow state, which would then cause exceptions
	- Handle the `UniqueImmutable` borrow
	- When borrowing, do a fake read to ensure the reference is valid (this breaks `kani::any_slice`, as it requires reading a symbolic slice)
	- Properly treat one-variant-ZST enums as ZSTs
	- Marginally better unsizing algorithm.
	  *random thought* 💭 when unsizing `T1` to `T2`, what if instead of trying to reconstruct a new `rust_val` with the modified pointer, i just wrote the value with its original type (`T1`) at offset 0 and then wrote the metadata at offset `size_of T1`?? Because at the low level this is exactly what happens I believe
	- Large arrays improvements:
		- Layout changes: instead of a `int Array.t` for `member_ofs`, I made a `FieldLayout` module, very similar to what Rust has internally: `Fields_shape.t = | None | Arbitrary of int Array.t | Array of int`, which is much more memory efficient for arrays since there is only one stride
		- Wherever possible, avoid casting `Z.t` from const generics into an `int`, since it may overflow!
		- This means we get less/no exceptions from very large arrays e.g. `[(); usize::MAX]`. We still can't run any code with these as we try generating the whole expression, but it's a start.
- Add some stubs to help Obol
- Reduce `(a * n) / n = a`
- Add configurable step and branch fuel via the CLI
- Add some additional information to a few errors


giltho:
> random thought 💭 when unsizing T1 to T2, what if instead of trying to reconstruct a new rust_val with the modified pointer, i just wrote the value with its original type (T1) at offset 0 and then wrote the metadata at offset size_of T1?? Because at the low level this is exactly what happens I believe

Hmm, is there a chance this would fail at catching a bug while re-creating the value? If not then sure

soteria-rust/lib/crate.ml
```diff
@@ -21,21 +21,21 @@ let as_fmt_env () =
...
-  let to_str = PrintUllbcAst.Ast.terminator_to_string (as_fmt_env ()) "" in
-  Fmt.pf ft "%s" (to_str term)
-
-let pp_fn_operand ft op =
-  let to_str = PrintUllbcAst.Ast.fn_operand_to_string (as_fmt_env ()) in
-  Fmt.pf ft "%s" (to_str op)
+let[@inline] mk_pp to_string =
+ fun ft v ->
+  let to_str = to_string (as_fmt_env ()) in
+  Fmt.pf ft "%s" (to_str v)
```
giltho, line 27: Nitpick style but
```ocaml
let mk_pp to_string = Fmt.of_to_string @@ to_string (as_fmt_env ())
```

giltho: Would be nice if Ullbc exposed pretty printers instead of to_string function. The issue with to_string is that you can't format them afterwards (e.g. break lines in the right place etc.)

soteria-rust/lib/error.ml
giltho, line 1: Niceee

soteria-rust/lib/layout.ml
```diff
@@ -1,25 +1,35 @@
...
 exception CantComputeLayout of string * Types.ty
-exception InvalidLayout
+exception InvalidLayout of Types.ty
+
+(** We use a custom type for the member offsets for layouts; this allows us to
+    use a more efficient representation for arrays [T; N], that doesn't require
+    N offsets. *)
+module Fields_shape = struct
+  type t =
+    | None  (** No fields present *)
```
giltho, line 16: Something something we have `field_shape option` at home

soteria-rust/lib/layout.ml
giltho, line 1: This is also a good thing I think! It's nice that we can be precise here.

soteria-rust/lib/layout.ml
giltho: (Can we call this `No_field` though to avoid clashes with Stdlib?)

soteria-rust/lib/layout.ml
```diff
@@ -167,62 +172,73 @@ let rec layout_of (ty : Types.ty) : layout =
...
   Session.get_or_compute_cached_layout_var variant @@ fun () ->
-  let discr_ty = enum_discr_ty adt_id in
-  let members = discr_ty :: field_tys variant.fields in
-  layout_of_members members
+  let variants = Crate.as_enum adt_id in
+  if
+    List.compare_length_with variants 1 = 0
+    && (List.for_all (fun ty -> (layout_of ty).size = 0)
+       @@ field_tys variant.fields)
+  then
```
giltho, line 231:
```ocaml
match variants with
| [ _single ] when (List.for_all (fun ty -> (layout_of ty).size = 0) @@ field_tys variant.fields) ->
    (* An enum with a single variant where all fields are ZST is a ZST *)
     ...
| _ -> ...
```

soteria-rust/lib/plugin.ml
```diff
@@ -181,6 +184,7 @@ let default =
           "-Zcrate-attr=feature(register_tool)";
           "-Zcrate-attr=register_tool(rusteriatool)";
           "--extern=rusteria";
+          "--edition=2024";
```
giltho, line 187: This should be up to the user?

soteria-rust/lib/state.ml
```diff
@@ -172,14 +173,16 @@ let with_ptr (ptr : Sptr.t) (st : t)
...
     let loc, ofs = Typed.Ptr.decompose ptr.ptr in
     let@ state = with_state st in
-    let++ v, state =
-      (SPmap.wrap (Freeable.wrap (fun st -> f (ofs, st)))) loc state
-    in
-    (v, state)
+    let* res = (SPmap.wrap (Freeable.wrap (fun st -> f (ofs, st)))) loc state in
+    match res with
+    | Missing _ as miss ->
+        if%sat Sptr.is_at_null_loc ptr then Result.error `UBDanglingPointer
```
giltho, line 183: Didn't we say we didn't like this?

soteria/lib/c_values/svalue.ml
```diff
@@ -671,6 +671,14 @@ let div v1 v2 =
...
   | _, _ when equal v2 one -> v1
   | Int i1, Int i2 -> int_z (Z.div i1 i2)
+  | Binop (Times, v, { node = { kind = Int i; _ }; _ }), Int j
+  | Int j, Binop (Times, v, { node = { kind = Int i; _ }; _ })
+    when Z.equal i j ->
+      v
+  | Binop (Times, { node = { kind = Int i; _ }; _ }, v), Int j
+  | Int j, Binop (Times, { node = { kind = Int i; _ }; _ }, v)
+    when Z.equal i j ->
+      v
```
giltho, line 681: I think part of the reduction project has to be about making this kind of thing more inspectable using a PPX

soteria/lib/soteria-std/seq.ml
```diff
@@ -11,3 +11,9 @@ let self_cross_product seq =
     | Cons (x, seq) -> append (map (fun y -> (x, y)) seq) (aux seq)
   in
   aux seq
+
+let rec init_aux_z f i j () =
+  if Z.lt i j then Cons (f i, init_aux_z f (Z.succ i) j) else Nil
```
giltho, line 16: Does this need to be exposed?

giltho (approved):

soteria-rust/lib/crate.ml
N1ark: yeah :/ it does expose pretty-printers but they're just the `deriving show` ones, whereas the `to_string` functions are for actual pretty-printing

N1ark: also i think what you put doesn't work because it will eval `as_fmt_env` which is an effect, which will crash

soteria-rust/lib/plugin.ml
N1ark: hmm yes, i guess.. Miri's default behaviour is to use edition 2024, I should just set it for the Miri plugin actually -- I'll eventually add a `soteria-rust ... -- -rustc-flag1 --rustc-flag2` option

PR #104, opened 2025-08-11
Iterate and substitute variables in rust_val

pcarrott:
Implement `iter_vars` and `subst` for `rust_val`. This allows the same functions to be implemented for serialized `Tree_block`s. The `Sptr` interface also requires these functions now.

### DIFF
--- a/soteria-rust/lib/sptr.ml
+++ b/soteria-rust/lib/sptr.ml
@@ -57,6 +57,9 @@ module type S = sig
 
   (** Get the allocation info for this pointer: its size and alignment *)
   val allocation_info : t -> T.sint Typed.t * T.nonzero Typed.t
+
+  val iter_vars : t -> (Svalue.Var.t * 'b ty -> unit) -> unit
+  val subst : (Svalue.Var.t -> Svalue.Var.t) -> t -> t
 end
 
 type arithptr_t = {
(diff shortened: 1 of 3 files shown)


N1ark (approved):
thanks!:)

soteria-rust/lib/sptr.ml
```diff
@@ -164,4 +167,15 @@ module ArithPtr : S with type t = arithptr_t = struct
 
   let as_id { ptr; _ } = Typed.cast @@ Typed.Ptr.loc ptr
   let allocation_info { size; align; _ } = (size, align)
+
+  let iter_vars { ptr; align; size; _ } f =
+    Typed.iter_vars ptr f;
+    Typed.iter_vars align f;
+    Typed.iter_vars size f
```
giltho, line 174:
```suggestion
  let iter_vars { ptr; align; size; tag = _ } f =
    Typed.iter_vars ptr f;
    Typed.iter_vars align f;
    Typed.iter_vars size f
```

Nitpick but for these kinds of situations I tend to avoid the `_` pattern because if some fields are added, `iter_vars` will be ignoring them silently leading to all kinds of bugs.
This is as opposed to e.g. `allocation_info` above, where the `_` is ok because

PR #110, opened 2025-08-15
Automatic publication of the odoc artifact

giltho:
- When a PR is created, the API is uploaded to `soteria-tools.github.io/soteria/pr-[number]`
   For instance, here: https://soteria-tools.github.io/soteria/api/pr-110/
- ~When a PR is opened, it is commented by the bot to get the link easily (though I could remove that really...)~
- ~When a PR is closed, the folder is overridden to delete that doc (I could also remove that really hahaha)~

Decided to only keep the publishing. I'll do the depublishing manually if needed


giltho:
After merging main, I can confirm that it updated successfully!

N1ark (approved):
niceee thank youuu

giltho:
I've slept on it and  I've decided that I'll remove automation on PR creation and PR deletion
It's not necessary, if we want I'll cleanup the pr folders manually, and it's just technical debt

N1ark:
> I've slept on it and I've decided that I'll remove automation on PR creation and PR deletion It's not necessary, if we want I'll cleanup the pr folders manually, and it's just technical debt

ah fair !

PR #111, opened 2025-08-15
Fix generated intrinsincs doc

giltho:
Before that, it would generate syntactically-invalid odoc.
Now, there is no more warnings, which is good because it allows to track when something goes wrong with odoc.


soteria-rust/scripts/intrinsics.py
```diff
@@ -148,11 +145,14 @@ def sanitize_var_name(name: str) -> str:
...
-            for attrib in fun["item_meta"]["attr_info"]["attributes"]
-            if "DocComment" in attrib
-        ]
+        doc = "\n".join(
+            [
+                attrib["DocComment"]
+                for attrib in fun["item_meta"]["attr_info"]["attributes"]
+                if "DocComment" in attrib
+            ]
+        )
```
N1ark, line 154:
```suggestion
        doc = "\n".join(
            attrib["DocComment"]
            for attrib in fun["item_meta"]["attr_info"]["attributes"]
            if "DocComment" in attrib
        )
```

N1ark (approved):
thanks!!

soteria-rust/scripts/intrinsics.py
giltho: Wait what
What's the type of this nightmare?

N1ark: ahahhaha it's an iterable

PR #112, opened 2025-08-15
Remove Make_seq, adjust documentation

giltho:
I removed Make_seq entirely and adjusted the tutorial accordingly.

I also took  the occasion to adjust signatures in the documentation. Specifically, I made sure that things of the type Symex are printed as `'a t` instead of `'a MONAD.t` which was a pain, and also that the horrible Fold signatures are written as-is.

Before:
image
image

After
image
image


N1ark (approved):
this is GREAT it was always such a painnnnn

PR #113, opened 2025-08-18
Soteria Stats

giltho:
Adds a module for statistics computation for Soteria. I tried to keep the API as simple as possible.

I still need to fix the tutorial that doesn't compile rn, but it's minor. I think you can start review, only `core_symex.mld` is going to change


soteria-c/test/cram/simple.t/run.t
```diff
@@ -76,7 +76,7 @@ Checking that fuel gets exhausted properly
     [Error: Failed assertion with trace
             [• Called from here: while_true.c:6:5-26;
              • Triggering operation: while_true.c:6:5-26]]
-  Executed 152 statements
+  Executed 153 statements
```
N1ark, line 79: were you counting wrong?:o

soteria/lib/symex/symex.ml
```diff
@@ -53,16 +79,42 @@ module type Base = sig
...
-  val run : fuel:Fuel_gauge.t -> 'a t -> ('a * sbool v list) list
+  (** [run ~fuel  p] actually performs symbolic execution of the symbolic
+      process [p] and returns:
+      - a list of obtained branches which capture the outcome together with a
+        path condition that is a list of boolean symbolic values.
+      - statistics correponding to the execution.
+
+      [fuel] corresponds to the {{!Fuel_gauge.t}fuel gauge} used for execution.
+  *)
+  val run : fuel:Fuel_gauge.t -> 'a t -> 'a run_result
```
N1ark, line 90: i was thinking the other day, maybe we can make the fuel gauge optional (and by default infinite)? just so that for small demos it looks cuter with the `~fuel` omitted

soteria/lib/symex/symex.ml
```diff
@@ -314,6 +383,7 @@ module Make (Sol : Solver.Mutable_incremental) :
                   L.debug (fun m ->
                       m "Exhausted branching fuel, not continuing")
               | Not_exhausted ->
+                  Stats.As_ctx.add_branches 1;
                   if is_sat (Solver.sat ()) then else_ () f
                   else L.trace (fun m -> m "Branch is not feasible"))
```
N1ark, line 388: maybe im being silly but to me this is ambiguous, as I would have expected that the explored branches is the branches that are actually executed; i would find it odd to have `stats.branches <> List.length (branches)` at the end of `Symex.run`
I think this is fine but maybe mention it in the documentation of Soteria.Stats that this is the number of attempted branches, including unfeasible ones? idk....

soteria/lib/symex/symex.ml
```diff
@@ -216,6 +284,7 @@ module Make (Sol : Solver.Mutable_incremental) :
   end
 
   let consume_fuel_steps n f =
+    Stats.As_ctx.add_steps n;
     match Fuel.consume_fuel_steps n with
     | Exhausted -> L.debug (fun m -> m "Exhausted step fuel")
     | Not_exhausted -> f ()
```
N1ark, line 290: ok so this answers my comment on the test; i don't think i agree with this way of counting? bc here this means if i have a step fuel of `10`, and run to exhaustion, the stats will tell me that i ran `11` steps which is not true
i would make the increment happen only in `Not_exhausted` here, but maybe you have your reasons

soteria/lib/symex/symex.ml
```diff
@@ -367,12 +440,26 @@ module Make (Sol : Solver.Mutable_incremental) :
         (with_section @@ fun () -> a () f);
         loop r
 
-  let run ~fuel iter =
+  type 'a run_result = {
+    stats : Stats.t;
+    results : ('a * Value.sbool Value.t list) list;
+  }
```
N1ark, line 446: \*in unison\* named tuples named tuples named tuples !!

soteria/lib/stats/soteria_stats.ml
```diff
@@ -0,0 +1,186 @@
...
+      let stats = get () in
+      f stats
+
+    let push_give_up_reason ~loc reason = wrap (push_give_up_reason ~loc reason)
+    let push_missing_without_fix reason = wrap (push_missing_without_fix reason)
+
+    let with_stats () f =
+      let stats = create () in
+      let res = try f () with effect Get, k -> Effect.Deep.continue k stats in
+      (res, stats)
```
N1ark, line 167: sorry this is insane levels of nitpick so feel free to ignore but i would much prefer having the `get` and `with_<effect>` functions close to each other to make it clear where the effect is raised 

```suggestion
    let with_stats () f =
      let stats = create () in
      let res = try f () with effect Get, k -> Effect.Deep.continue k stats in
      (res, stats)

    let wrap f =
      let stats = get () in
      f stats

    let push_give_up_reason ~loc reason = wrap (push_give_up_reason ~loc reason)
    let push_missing_without_fix reason = wrap (push_missing_without_fix reason)
```

N1ark: (the diff isn't clear, but I just moved `with-stats` before `wrap`/`push_give_up_reason`/`push_missing_without_fix`

N1ark (approved):
lgtm !! this is really neat, excited to try it 
ill probably make Rusteria using it a separate PR for the sake of clarity and so we can get this merged

soteria-c/lib/soteria_c_lsp.ml
```diff
@@ -11,10 +11,16 @@ let cerb_loc_to_range loc =
...
+let get_abort_diagnostics (stats : Csymex.Stats.t) =
   let open Syntaxes.List in
-  let+ msg, loc = Csymex.flush_give_up () in
+  let list_reasons =
+    stats.give_up_reasons
+    |> Hashtbl.Hstring.to_seq
+    |> Seq.concat_map (fun (reason, locs) ->
+           Seq.map (fun loc -> (reason, loc)) (Dynarray.to_seq locs))
+    |> List.of_seq
+  in
```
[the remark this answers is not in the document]
giltho: It's more that Hashtbl doesn't expose a way of getting a list of bindings, only a sequence.
But yes, supposedly it's more efficient anyway. A list has more chance to end up in the major heap

soteria-c/test/cram/simple.t/run.t
giltho: No, it's just that previously I was calling it like:
```ocaml
Csymex.consume_step_fuel 1;
Stats.increment_stmt_counter ();
```
and now `consume_step_fuel` itself is incrementing. That specific test exhausts step fuel, so previously the stat wasn't incremented at the end but now it is.

soteria/lib/stats/soteria_stats.ml
giltho: Applied this! Though `wrap` doesn't exist anymore, I changed to `apply`.

giltho:
Ok I believe that I have applied all comments.
In addition, as said on discord:
- Reworked the API, which now has `run`, `run_with_stats`, `run_needs_stats`.
- I've extracted a notion of `with_stats`, which allows for nicer code
- I've modified stats effect from `Get` to `Apply f`, and a handler that just ignores the function to apply (if the user doesn't want to pay the cost)

N1ark (approved):
thanks!!

PR #115, opened 2025-08-21
Rust work

N1ark:
I can split this if needed; most of the diff should be in the demo code and test scripts.
- Add intrinsics `cctz`, `cttz_nonzero`
-


soteria-rust/lib/rustsymex.ml
```diff
@@ -31,22 +40,13 @@ let with_loc_immediate ~loc f =
...
-
 let[@inline] with_loc_err () f =
   let loc = get_loc () in
   Result.map_error (f ()) (fun e -> (e, loc))
 
 let error e = Result.error (e, get_loc ())
-let of_opt = function Some x -> return x | None -> vanish ()
-let of_opt_not_impl ~msg = function Some x -> return x | None -> not_impl msg
+let not_impl msg = give_up ~loc:(get_loc ()) msg
+let of_opt_not_impl msg = some_or_give_up ~loc:(get_loc ()) msg
```
giltho, line 49: Ah ignore previous comment then

soteria-rust/scripts/config.py
```diff
@@ -56,7 +59,38 @@ def filter_tests(opts: CliOpts, tests: Iterable[Path]) -> list[Path]:
...
+    with open(file, "r") as f:
+        for line in f:
+            if line.startswith(prefix):
+                return line.rstrip("\n").split(prefix)[1].strip()
+    return None
+
+
 def kani(opts: CliOpts) -> TestConfig:
+
+    @with_cache
```
giltho, line 87: Python is so elegant. I'd give a lot to have
```ocaml
@with_cache
let f x = ...
```

Instead of
```ocaml
let f x =
   let@ () = with_cache in 
   ...
```

You could almost make it a trivial ppx, but then even that syntax is ugly it'd be:
```ocaml
let%with_cache f x = ...
(* or *)
let f x =
   ...
   ...
   [@@ with_cache]
```

soteria/lib/symex/symex.ml
```diff
@@ -306,6 +308,11 @@ Extend (struct
     in
     aux [] learned
 
+  let solver_sat () =
```
giltho, line 311:
```ocaml
module Solver = struct 
   include Solver
   
   let sat () = ...
end
```
to prevent future misuses of `Solver.sat`?

giltho (approved):

soteria-rust/scripts/config.py
N1ark: yeah it's unfortunate, i wish we had prefix syntax/ppx decorators; also the `let%with_cache` wouldn't even be as nice bc you cant stack ppx derives like you stack decorators </3

N1ark:
@giltho should be good -- can you just double check 646ec4a

giltho:
lgtm :)

PR #120, opened 2025-09-01
Add `Soteria.Solvers`,  `Z3` as a functor

N1ark:
Add `Soteria.Solvers`, with `Z3`! Though tbh I called it `Z3` but it doesn't have much Z3-specific behaviour, its just because its from`Z3_exe` and it uses the config for Z3's location.

It's a functor that receives a module `Value`, defined as:
```ocaml
module type Value = sig
  type t
  type ty

  (** A list of commands that need to be aknowledged on start up of the solver.
      These are also sent everytime [Solver_interface.S.reset] is called. *)
  val init_commands : sexp list

  (** Encode a type into a SMTLib sort *)
  val sort_of_ty : ty -> sexp

  (** Encode a value into a SMTLib value *)
  val encode_value : t -> sexp
end
```

It was actually super straightforward to change! A next step would probably be to factor a few things out of `C_solver`, to allow it to be reused as well.... but I'll leave this for later.

This also nicely fits potential future plans to generate the encoding!

Also did a drive-by change to add a `simplify` to `Analysis`, and use it with `Interval`!
Evaluating this with Kani, on 5 iterations: went from 17.017s to 16.813s (-1.2%)


giltho (approved):
Very nice! I was a bit afraid of functor hell when looking at this PR, but it still looks like the overall API doesn't change so it's nice.

PR #121, opened 2025-09-02
Add `Bv_values`

N1ark:
Add a modules `Bv_values`, that is entirely backed by bitvectors.
I honestly didn't know how to split this lmao; it's originally a copy of `C_values` that I modified to remove integers, and to which i added a lot more bit vector reductions. `Bv_solver` is also just a copy of `C_solver`. 

The only actual relevant file for you is probably `svalue.ml`.

Also while doing this I realised the sign of the integer is relevant when doing float-int conversions so I just added that on the `C_values` side, and patched the interpreters.

Given bitvectors have wrapping and also that there are both signed and unsigned comparison operators, I for now don't port the `Interval` analysis to it; I have it done but I'll push it in a separate PR in case we want to discuss it separately, + i need to evaluate it anyways.


giltho (approved):
Not much to say, there's a lot but it looks ok and exciting.

I'm just curious to see how it fares in general!

PR #122, opened 2025-09-02
Extend Typed for Rust, prepare migration

N1ark:
Update Rusteria to use `Bv_values`! This is the last big PR, with all my recent work. Whatever I do next will be on separate and hopefully smaller branches.

Again I tried organising it in a way that commits can be read individually; the first couple commits only prepare stuff and don't really compile, but it felt relevant to separate them. Everything else after compiles and works :)

- Use `Bv_values` to Rusteria. `Typed` is extended quite a bit to make this nicer and easier, e.g. `Typed.BitVec.usize x` is `Typed.BitVec.mk_masked (8 * Crate.pointer_size ()) x`.
  I'm very happy with this change; almost everywhere in the interpreter I now do typed casts that check for the size of values, so I can be much more confident that what I'm writing/reading is correct.
  This also made some intrinsics (`ctpop`, `cttz`...) and transmuting/splitting much easier, as I got rid of a bunch of ugly `(x / 2^n) % 2^m` and replaced these with `BitVec.extract`.
- Handle enum niches! This was simpler than I expected, and is done mostly between `Layout` and `Encoder`. An `enum` has a layout consisting of a `Tag_layout` and a `Fields_shape` for each variant.
  The `Tag_layout` stores the tag's type, offset, encoding (`Direct` if it is prepended to the fields, `Niche` if it is actually not written separately but rather over the fields), and the tag for each variant (except untagged or uninhabited variants)
	```ocaml
	module Tag_layout = struct
	  type encoding = Direct | Niche of Types.variant_id
	  and t = {
	    offset : int;
	    ty : Types.literal_type;
	    encoding : encoding;
	    tags : Z.t option Array.t;
	  }
	end
	
	module Fields_shape = struct
	  type t =
	    | Primitive
	    | Arbitrary of int Array.t
	    | Enum of Tag_layout.t * t Array.t
	    | Array of int
	end
	```
- Small driver printing changes, better error messages for `UBTransmute`, and small script changes


giltho (approved):
A bit difficult to review, overall I just have tiny style comments.

Also, please do in a subsequent PR:
- Improved syntax for integers (e.g. U32.(128s))
- Attempt at typing operations with `Overflowed` and `NotOverflowed` or something to try and avoid invalid operations

PR #125, opened 2025-09-05
Rusteria syntactic sugar and typed constraints

N1ark:

PR #131, opened 2025-09-10
Symbolic abstractions

giltho:
This PR introduces the `Sym_data` folder, which contains utilities for what I call (in my thesis) Symbolic Abstractions. Really... simple abstract domains.

S_eq is an abstraction that implements equality, while S_int is an abstraction over integers (i.e. a symbolic "thing" that can be interpreted as an integer).
I also add S_range (used for TreeBlocks) and S_map (used for PMap) -- We really need a linter to enforce naming conventions 😭 

I also extract more clearly what `sbool` is and explicitly, i.e. an abstraction over booleans. Some of the resulting code is slightly more verbose but I think that's ok.

Also, some of the code in TreeBlock is a bit verbose because of `MemVal.S_int.zero ()` everywhere but #125  fixes this already :)


N1ark (commented):
Conceptually this is cool but I am not very convinced, because rn it adds a lot of code and signatures that are barely used, even where they seem to fit (`Symex`).

The dependency `Symex` -> `Sym_data` is quite unpractical, and I wonder if there is not a cleverer split to do, like:
1. `Sym_val` for symbolic values, ie. just defining the shape of different sorts of values (with `S_bool`, `S_int`, `S_eq`, `S_elt`, `S_range`)
2. `Symex`, defines symbolic processes using symbolic values
3. `Sym_data`, defines data structures using symbolic values processes. `s_map` right now, but could include sets, maybe binary trees, idk.

We could just re-export `Sym_val` from `Sym_data` to only provide one module of symbolic things, but at least it would avoid the weird split we have rn, because `Symex` not reusing `S_eq` etc. is super weird to me

soteria/lib/sym_data/s_map.ml
```diff
@@ -0,0 +1,124 @@
...
+      In both cases, the resulting [key'] can be used to insert or update the
+      map directly without needing to search for it again. *)
+  val find_opt : key -> 'a t -> (key * 'a option) Symex.t
+end
+
+module Make (Symex : Symex.S) (Key : Key.S with module Symex = Symex) :
+  S with type key = Key.t and module Symex = Symex = struct
+  module Symex = Symex
+  open Symex.Syntax
+  module Raw_map = Stdlib.Map.Make (Key)
```
N1ark, line 52: looking back -- why don't you `include Raw_map`? or at least reexpose some of it, since e.g. one probably wants `add`, `empty` functions

N1ark (changes requested):
still not convinced tbh :/ (sorry if this wasnt ready for review)

PR #135, opened 2025-09-18
cleanup alcotest tests

giltho:
Nothing big, just splitting the alcotest tests for soteria_std in more files, and use helper functions to register the tests


soteria/tests/soteria_std/graph_tests.ml
```diff
@@ -0,0 +1,113 @@
+open Test_register
+
+let register f = register "Graph" f
```
N1ark, line 3: misleading because `f` isn't a function but the name of the test' with the function curried behind... so maybe just do `let register = register "Graph"` and curry it all

N1ark (approved):
lgtm! nit

PR #138, opened 2025-09-26
Fix small unsoundness

giltho:
Assume doesn't currently emit a sat check (to avoid running a billion sat checks when we assume a lot of stuff during e.g. production.

It does mean we could finish execution with branches that have left-over unchecked path conditions which could be unsat.

Thankfully, we keep track of what part of the state has been sat checked already, so we just have to run one last sat check at the end and for these corner cases it will emit a sat check and for the others it won't


giltho:
I had it in the back of my mind but it's really @pcarrott who spotted it :)

PR #140, opened 2025-10-02
Update README for Rust

N1ark:


giltho (approved):
I think we need a rewrite of the whole thing anyway, but this a very good improvement already

PR #141, opened 2025-10-05
Bump Charon, resolve trait consts, fix float-int casting

N1ark:
Everyone's favourite PR, a Charon bump :) 

- Properly resolve trait consts in `resolve_constant`, for the simple case (which is the most common)
- Extract `transmute_literal` from `transmute`, for the simpler literal cases.
- Make `Bv_values.BV.{of_float,to_float}`, and `C_values.Float.{of_int,to_int}` receive a rounding mode! It uses to default to "nearest ties to away", which is not correct (at least for Rust, and probably for C too).
- Implement `NullaryOp.OffsetOf` -- it wasn't being reached in any tests I found but it still exists so might as well
- Bump Charon; the main change is that a variant's discriminant is now a literal (ie. can be a `char`/`bool`, mainly) rather than a int/uint scalar. Also they renamed a buuunch of types from `_raw` to `_kind` which is nice.


giltho (approved):
nice

PR #147, opened 2025-10-20
Existential quantifiers as symbolic values

pcarrott:
Added a new symbolic value `Exists of (Var.t * ty) list * t` to `bv_values`. This is needed for RUXt.

Can do the same for `c_values` for consistency, though it is not something I need atm.


soteria/lib/bv_values/svalue.ml
```diff
@@ -346,6 +360,14 @@ let rec subst subst_var sv =
       let e' = subst subst_var e in
       if equal c c' && equal t t' && equal e e' then sv
       else Ite (c', t', e') <| sv.node.ty
+  | Exists (vs, v) ->
+      let subst_var v =
+        match List.find_opt (fun (v', _) -> Var.equal v v') vs with
+        | Some (v, _) -> v
+        | None -> subst_var v
```
```suggestion
        if List.exists (fun (v', _) -> Var.equal v v') vs then v
        else subst_var v
```

soteria/lib/bv_values/eval.ml
```diff
@@ -48,6 +48,14 @@ let eval_unop : Unop.t -> t -> t = function
   | FRound rm -> Float.round rm
 
 let rec eval ~eval_var (x : t) : t =
+  let eval_without_vars vs =
+    let eval_var v ty =
+      match List.find_opt (fun (v', _) -> Var.equal v v') vs with
+      | Some (v, ty) -> mk_var v ty
+      | None -> eval_var v ty
+    in
+    eval ~eval_var
```
N1ark: could you move this into the `Exists` branch since it's only used there? this also avoids an intermediary function since the inputs are known

soteria/lib/bv_values/encoding.ml
```diff
@@ -107,6 +108,10 @@ let rec encode_value (v : Svalue.t) =
           List.map (fun v -> seq_singl (encode_value_memo v)) vs |> seq_concat)
   | Ite (c, t, e) ->
       ite (encode_value_memo c) (encode_value_memo t) (encode_value_memo e)
+  | Exists (vs, sv) ->
+      let exists l x = list [ atom "exists"; list l; x ] in
+      let encode_var_memo (v, ty) = list [ var v; sort_of_ty ty ] in
+      exists (List.map encode_var_memo vs) (encode_value_memo sv)
```
N1ark: 1. I avoid direct uses of `atom` in encodings, as it can be error prone and overall not super nice; could you move this to `solvers/smt_utils`?
  Also ideally use the appropriate Smtlib functions, so i think it should rather look like
    ```ocaml
    let exists qs e = app_ "exists" [ list qs; e ]
    ```
2. `encode_var_memo` seems to imply there is some memoisation happening, which there isn't; can you rename it like `encode_binding` please :)

N1ark (commented):
Nice addition thanks :) 
Sorry ton of nitpicks just to keep the codebase clean

giltho:
@N1ark This is ready for another round of reviews :)

soteria/lib/bv_values/svalue.ml
```diff
@@ -685,6 +701,49 @@ module rec Bool : Bool = struct
...
+      List.filter (fun (v, _) -> Var.Hashset.mem body_vars v) binders
+    in
+    match binders with [] -> body | _ -> Exists (binders, body) <| TBool
+
+  (** * [exists_n ~not_in tys mk] creates an existential with [length tys]
+      variables of types [tys], that are not in [not_in], and with body created
+      by [mk : t list -> t] which takes the created variables as input in the
+      same order as [tys]. *)
+  let exists_n ~not_in tys mk =
+    (* FIXME: Ideally, the not_in parameter would not be necessary. What we
```
N1ark, line 717: i'm ok with this, but it feels a bit ugly; what if `mk_exists` just wasn't implemented here and was implemented separately? 
e.g. have a `logic.ml` file in `bv_values`, with a `Make (Symex: Symex.Base with Symex.Value = Svalue)`, and then use `fresh_var` in there? if you really need this merged asap go for it but it eeks me a bit

soteria/lib/bv_values/typed.mli
```diff
@@ -135,6 +135,22 @@ module type Bool_ := sig
   val not : [< sbool ] t -> [> sbool ] t
   val distinct : 'a t list -> [> sbool ] t
   val ite : [< sbool ] t -> 'a t -> 'a t -> 'a t
+  val exists_1 : not_in:_ t -> 'a ty -> ('a t -> [< sbool ] t) -> [> sbool ] t
```
N1ark, line 138: TIL `_ t` is valid syntax

soteria/lib/bv_values/svalue.ml
N1ark: alternatively, does it actually matter whether variables clash within the existential? if we handle it properly it should be fine no? or does e.g. smtlib forbid it?

N1ark (approved):
lgtm; im being a bit annoying about the constructor, but it's ok as is

giltho (approved):

PR #150, opened 2025-10-27
Add interval analysis to `Bv_values`

N1ark:
- Add an interval analysis strongly inspired by https://ceur-ws.org/Vol-1617/paper8.pdf
- Add a `checked` flag to `Add` and `Sub` to allow some reductions around inequalities with them; e.g. while `c1 + x < c2 <=> x < c2 - c1` is generally not true for bitvectors with wrapping semantics, if the addition is known to not overflow then this is sound. This is essential to simplify inequalities, which the interval analysis relies on, which in turn TreeBlock relies on.
- Add/fix some reductions

### DIFF
--- a/soteria-rust/test/cram/kani.t/any_vec.rs
+++ b/soteria-rust/test/cram/kani.t/any_vec.rs
@@ -1,0 +1,8 @@
+// Small test for kani::vec::any_vec, which can trigger errors
+// if we don't do pointer offsets properly
+
+#[kani::proof]
+fn len_capacity_invariant() {
+    let v = kani::vec::any_vec::<i32, 16>();
+    assert!(v.capacity() >= v.len());
+}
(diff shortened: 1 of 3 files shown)


soteria/lib/bv_values/analyses.ml
```diff
@@ -50,3 +60,431 @@ module None : S = struct
   let add_constraint () v = (v, Var.Set.empty)
   let encode ?vars:_ () = Iter.empty
 end
+
+module Interval : S = struct
+  (* we only include stuff from Z we want  *)
```
giltho:
```suggestion
  (** An interval analysis for bitvectors inspired by
       https://ceur-ws.org/Vol-1617/paper8.pdf *)


  (* we only include stuff from Z we want  *)
```

giltho:
Let's merge?

N1ark:
> Let's merge?

yes sorry i was waiting for CI and got distracted

PR #172, opened 2025-11-11
Add equality analysis

N1ark:
Add an equality analysis, that will remember known equalities using a union find, and substitute any matching expression with the cheapest version of that expression.
E.g. if it knows `X = Y / 8`, it will substitute any new occurrences of `Y / 8` with `X`.

There is a cost function to determine this; it is relatively arbitrary and based on personal superstition
```ocaml
 let rec cost (v : Svalue.t) : int =
    match v.node.kind with
    | Binop (op, l, r) -> cost_binop op + cost l + cost r
    | Unop (op, v) -> cost_unop op + cost v
    | Ite (i, t, e) -> cost i + cost t + cost e
    | Nop (_, vs) -> costs vs
    | Var _ -> 3
    | Float _ -> 2
    | Seq vs -> costs vs
    | Ptr _ | Bool _ | BitVec _ -> 1

  and costs vs = List.fold_left (fun acc v -> acc + cost v) 0 vs

  and cost_binop : Svalue.Binop.t -> int = function
    | FAdd | FSub | FMul | FDiv | FEq | FLt | FLeq -> 3
    | _ -> 1

  and cost_unop : Svalue.Unop.t -> int = function
    | FRound _ | FIs _ -> 5
    | BvOfFloat _ | FloatOfBv _ -> 4
    | _ -> 1
```

### DIFF
--- a/soteria/lib/bv_values/analyses.ml
+++ b/soteria/lib/bv_values/analyses.ml
@@ -502,6 +502,11 @@ module Equality : S = struct
     let default = (UnionFind.new_store (), IMap.empty)
   end)
 
+  (* override to account for UnionFind  *)
+  let save d =
+    let uf, st = Dynarray.get_last d in
+    Dynarray.add_last d (UnionFind.copy uf, st)
+
   let rec cost (v : Svalue.t) : int =
     match v.node.kind with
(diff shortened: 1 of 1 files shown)


giltho:
It *will* cause a billion conflicts but we have ✨time✨

Issue #177, opened 2025-11-17
Provenance options


N1ark:
Currently, we only run a somewhat flawed version of [strict provenance](https://doc.rust-lang.org/nightly/std/ptr/index.html#provenance): we allow ptr-int casts, which lose provenance, but we also allow int-ptr casts, which create a dangling reference at the decayed address, with no provenance.
We should instead opt to do something like [what Miri does](https://github.com/rust-lang/miri#miri--z-flags-and-environment-variables), by adding a `--provenance` flag:
- `--provenance=strict` preserves the current behaviour, with additionally int-ptr casts being UB
- `--provenance=permissive` enables permissive provenance, where `int-ptr` casts angelically pick the right provenance if possible.
  Conveniently, the `DecayMap` already stores what addresses were exposed ! So we "just" need to reverse that match and see if any address matches, somewhat.

Optionally we could do like Miri: default to `permissive` with a warning, with an explicit `permissive` flag erasing that warning. Or make the default something like `permissive-loud`, which issues those warnings.

PR #185, opened 2025-11-22
Add provenance flag!

N1ark:
Fixes https://github.com/soteria-tools/soteria/issues/177

- Add a `--provenance` flag, with possible values `provenance` (default), and `strict`.
  Permissive provenance allows the following snippet to run:
  ```rust
	fn main() {
	    let mut x: u8 = 0;
	    let p = &mut x as *mut u8;
	    let p_int = p.expose_provenance();
	    let p_back = std::ptr::with_exposed_provenance::<u8>(p_int) as *mut u8;
	    unsafe {
	        *p_back = 1;
	        assert!(*p == 1);
	    }
	}
	```
- Add meta information to `State`, which tracks for each allocation it's size, alignment, the span where's it's allocated, the allocation kind (similar to what is done in https://github.com/rust-lang/rust/blob/be0ade2b602bdfe37a3cc259fcc79e8624dcba94/compiler/rustc_middle/src/mir/interpret/mod.rs#L260-L276) and the root of the tree borrows tree (unused for now)
- Properly disable all read/write access to function allocations

### DIFF
--- a/demo/darpa/simple.rs
+++ b/demo/darpa/simple.rs
@@ -3,7 +3,6 @@
 
 /// A classic overflow error when checking before adding two numbers
 #[kani::proof]
-#[kani::should_panic]
 fn overflow() -> u32 {
     let a: u32 = kani::any();
     let b: u32 = kani::any();
@@ -28,7 +27,6 @@ fn overflow_fixed() -> u32 {
 
 /// Soteria Rust can detect memory leaks, caused by dynamically allocated memory
 #[kani::proof]
-#[kani::should_panic]
 fn memory_leak() {
     let allocated = Box::new(11);
     std::mem::forget(allocated);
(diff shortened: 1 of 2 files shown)


soteria-rust/lib/sptr.ml
```diff
@@ -32,33 +53,62 @@ module DecayMap : DecayMapS = struct
...
       in
       Soteria.Symex.Compo_res.get_ok res
+
+  let from_exposed (loc_int : [< sint ] Typed.t) st =
+    (* UX: we only consider the first one; this is more or less correct, as per the
+       documentation of [with_exposed_provenance]: "The provenance of the returned pointer is
+       that of some pointer that was previously exposed"
+
+       See https://doc.rust-lang.org/nightly/std/ptr/fn.with_exposed_provenance.html
+       *)
```
giltho: I think this is fine really.

PR #186, opened 2025-11-22
Simplify State interface and printing

giltho:
- I realised that `load`/`store` were not really used and that only `load_aggregate`/`store_aggregate` were used.
- Improve printing using print_ast instead of print_program


N1ark (approved):
hehe this is what soteria-rust does too!

PR #188, opened 2025-11-23
Don't re-expose symex in every abstraction

giltho:
I realised instead of having functors of the form:
```ocaml
module Make (Symex: Symex.Base) (Param: ParamS with module Symex = Symex)
```
we could just make the `ParamS` signatures parametric on Symex directly.

I find this much nicer, and I believe this will make the abstraction PR much nicer as it'll be much easier to just re-use existing bits.


N1ark:
should we merge this?

PR #190, opened 2025-11-25
docs(soteria/lib/soteria_std): add module documentation and unified entry point

luthenwald:
- adds synopsis to all modules in `soteria/lib/soteria_std/`.
- adds documentations to `list.ml`, `monad.ml` & `reversible.ml`.
- creates `soteria_std.ml` that re-exports all modules.

### DIFF
--- a/soteria/lib/soteria_std/cmdliner_helpers.ml
+++ b/soteria/lib/soteria_std/cmdliner_helpers.ml
@@ -5,7 +5,7 @@
 let file_as_absolute =
   let pp = Format.pp_print_string in
   let parse path =
-    let path = Path.normalise path in
+    let path = Filename.normalise path in
     if Sys.file_exists path then Ok path
     else Error ("No such file or directory: " ^ path)
   in
@@ -16,7 +16,7 @@ let file_as_absolute =
 let dir_as_absolute =
   let pp = Format.pp_print_string in
   let parse path =
-    let path = Path.normalise path in
+    let path = Filename.normalise path in
     if Sys.file_exists path then
       if Sys.is_directory path then Ok path
       else Error ("Not a directory " ^ path)
--- a/soteria/lib/soteria_std/soteria_std.ml
+++ b/soteria/lib/soteria_std/soteria_std.ml
@@ -3,6 +3,7 @@
 module Bimap = Bimap
 module Cmdliner_helpers = Cmdliner_helpers
 module Dynarray = Dynarray
+module Filename = Filename
 module Foldable = Foldable
 module Graph = Graph
 module Hashset = Hashset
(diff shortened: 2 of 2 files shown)


soteria/lib/soteria_std/map.ml
```diff
@@ -1,3 +1,8 @@
+(** A wrapper around [Stdlib.Map] that includes a pretty-printer.
```
giltho: Let's be consistent between modules.
`list.ml` says "extensions of Stdlib.List` while this says "a wrapper around Stdlib.Map`

I prefer extensions

soteria/lib/soteria_std/path.ml
```diff
@@ -1,3 +1,5 @@
+(** Path normalization utilities. *)
```
giltho: Realising that this module should be implemented as an extension of `Filename`

soteria/lib/soteria_std/list.ml
```diff
@@ -16,9 +22,11 @@ let take_count n l =
   let result = aux n l in
   (result, !taken)
 
+(** Combine two lists into a list of pairs. *)
 let combine_opt l1 l2 =
   try Some (combine l1 l2) with Invalid_argument _ -> None
 
+(** Partition a list of results into successful values and errors. *)
```
giltho: This is not what the function does I believe. It was correct in your previous commits though

soteria/lib/soteria_std/option.ml
```diff
@@ -1,14 +1,23 @@
+(** Extensions to [Stdlib.Option] with monadic operations. *)
```
giltho: There are more than monadic operations I believe

soteria/lib/soteria_std/path.ml
```diff
@@ -1,3 +1,7 @@
+(** Extensions to [Stdlib.Filename] with utility functions. *)
```
giltho: But then the file itself should be renamed filename.ml

soteria/lib/soteria_std/seq.ml
```diff
@@ -12,6 +14,7 @@ let self_cross_product seq =
   in
   aux seq
 
+(** Create a sequence of values generated by a function up to a given bound. *)
```
giltho: Maybe emphasise that the difference with Stdlib.Seq.init is that the input parameter is a Z.t instead of an int

giltho (commented):
Some more minor corrections :)

giltho (approved):
Thank you!

If you're happy for us to merge could you please change the status of the PR from Draft to Ready for review, and confirm that you are happy with our [Contributor License Agreement](https://github.com/soteria-tools/soteria/blob/30f093661ed696de9f6540206e0175487cf3c0fc/CLA.md)? (It just says that your contributions are done under Apache2.0 license that we can reuse as part of this open source project)

PR #198, opened 2025-12-10
Use `State_monad` everywhere

N1ark:
Rather than passing `State_intf` everywhere and having these functors re-instantiate `State_monad`, `Encoder`, etc + re-lifting everything from `Sptr`, `Encoder` etc, we re-expose the lifted version of these modules in `State_monad`, and pass that monad everywhere. 

The biggest diff is in `State_monad` since of course we need to define the signature and re-expose some stuff, but I think it makes basically everything else cleaner, and makes it much clearer to see what depends on the state, which I like a lot. This also means we can get rid of the ugly `let^^` (lift `Symex.Result` by wrapping the error), `let$` (lift `DecayMapMonad`) and `let$$` (lift `DecayMapMonad.Result`).

### DIFF
--- a/soteria-rust/lib/builtins/alloc.ml
+++ b/soteria-rust/lib/builtins/alloc.ml
@@ -2,8 +2,8 @@ open Rust_val
 open Typed.Infix
 open Typed.Syntax
 
-module M (State_monad : State_monad.S) = struct
-  open State_monad
+module M (Rust_state_m : Rust_state_m.S) = struct
+  open Rust_state_m
   open Syntax
 
   let alloc ?(zeroed = false) args =
--- a/soteria-rust/lib/builtins/intrinsics_impl.ml
+++ b/soteria-rust/lib/builtins/intrinsics_impl.ml
@@ -5,15 +5,15 @@ open Typed.Syntax
 open Typed.Infix
 open Rust_val
 
-module M (State_monad : State_monad.S) : Intrinsics_intf.M(State_monad).Impl =
-struct
-  include Intrinsics_stubs.M (State_monad)
-  module Core = Core.M (State_monad)
-  open State_monad
+module M (Rust_state_m : Rust_state_m.S) :
+  Intrinsics_intf.M(Rust_state_m).Impl = struct
+  include Intrinsics_stubs.M (Rust_state_m)
+  module Core = Core.M (Rust_state_m)
+  open Rust_state_m
   open Syntax
 
   (* some utils *)
-  type 'a ret = ('a, unit) State_monad.t
+  type 'a ret = ('a, unit) Rust_state_m.t
 
   (* we retype these to avoid non-generalisable type variables in ['a Rust_val.t] *)
   let[@inline] as_ptr (v : rust_val) =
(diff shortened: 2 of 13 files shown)


soteria-rust/lib/builtins/intrinsics_impl.ml
```diff
@@ -208,10 +207,10 @@ module M (State : State_intf.S) : Intrinsics_intf.M(State).Impl = struct
...
       [`StdErr (name ^ " overlapped")]. *)
   let check_overlap name l r size =
-    let^^ l_end = Sptr.offset ~signed:false l size in
-    let^^ r_end = Sptr.offset ~signed:false r size in
-    let$ dist1 = Sptr.distance l r_end in
-    let$ dist2 = Sptr.distance r l_end in
+    let* l_end = Sptr.offset ~signed:false l size in
+    let* r_end = Sptr.offset ~signed:false r size in
+    let* dist1 = Sptr.distance l r_end in
+    let* dist2 = Sptr.distance r l_end in
```
giltho: ABSOLUTELY

soteria-rust/lib/builtins/intrinsics_impl.ml
```diff
@@ -496,7 +495,7 @@ module M (State : State_intf.S) : Intrinsics_intf.M(State).Impl = struct
   let maxnumf128 ~x ~y = float_minmax ~is_min:false ~x ~y
 
   let ptr_guaranteed_cmp ~t:_ ~ptr ~other =
-    State.with_decay_map_res @@ Core.eval_ptr_binop Eq (Ptr ptr) (Ptr other)
+    Core.eval_ptr_binop Eq (Ptr ptr) (Ptr other)
```
giltho: F*** YEAH

soteria-rust/lib/state_monad.ml
```diff
@@ -195,17 +391,36 @@ module Make (State : State_intf.S) = struct
...
     let ( let* ) = bind
     let ( let+ ) = map
     let ( let^ ) x f = bind (lift_symex x) f
     let ( let^+ ) x f = map (lift_symex x) f
-    let ( let^^ ) x f = bind (State.lift_err x) f
-    let ( let^^+ ) x f = map (State.lift_err x) f
-    let ( let$ ) x f = bind (State.with_decay_map x) f
-    let ( let$+ ) x f = map (State.with_decay_map x) f
-    let ( let$$ ) x f = bind (State.with_decay_map_res x) f
-    let ( let$$+ ) x f = map (State.with_decay_map_res x) f
```
giltho: <3

soteria-rust/lib/builtins/alloc.ml
```diff
@@ -2,8 +2,8 @@ open Rust_val
 open Typed.Infix
 open Typed.Syntax
 
-module M (State : State_intf.S) = struct
-  open State_monad.Make (State)
+module M (State_monad : State_monad.S) = struct
```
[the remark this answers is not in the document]
N1ark: hmm `Rust_state_m`?...

PR #202, opened 2025-12-12
Make layouts symbolics

N1ark:
- Make all information in layouts *symbolic*: the size, alignment, and field offsets. This makes the consumer code for the layouts slightly nicer (since we're not casting integers/`Z`s to `Typed` all the time), but also some stuff more annoying; e.g. `Layout.layout_of` returns a `Rustsymex.Result`, for the error case where a layout is invalid (namely, too large).

  For bi-abduction we'll probably want to assume layouts are valid because reporting those errors is noisy and unnecessary, but there are cases where concrete layouts are invalid and we should signal those. This used to be a raised exception but that was icky.

- Properly implement `transmute` and `assert_zero_valid`, remove `Encoder.transmute`, `Encoder.split`!
  
  I had weird hacks to implement those but really what we should do is just take the state, and write then read from it to transmute. This is now what happens! Added a `Core.transmute` function that does exactly that (and likewise for `assert_zero_valid`, we make a zeroed allocation and try reading it).
  This buffes up a bit `Rtree_block`, but it makes sense and is much cleaner. 

Next step is cleaning up the layout + encoding story; the encoder shouldn't need to match on type × value, but rather on layout(type) × value. I also want to throw away or better isolate layout computation. I'll do this in a follow up PR.

### DIFF
--- a/soteria/lib/bv_values/typed.mli
+++ b/soteria/lib/bv_values/typed.mli
@@ -99,11 +99,21 @@ val v_false : [> sbool ] t
 val bool : bool -> [> sbool ] t
 val as_bool : 'a t -> bool option
 val and_ : [< sbool ] t -> [< sbool ] t -> [> sbool ] t
+
+(** Similar to [and_], but the rhs is only evaluated if the lhs is not the
+    concrete false. In other words, this is a short-circuiting and. Avoids some
+    errors, like a division by zero in [0 != x && n / x] when [x] is [0]. *)
 val and_lazy : [< sbool ] t -> (unit -> [< sbool ] t) -> [> sbool ] t
+
 val conj : [< sbool ] t list -> [> sbool ] t
 val split_ands : [< sbool ] t -> ([> sbool ] t -> unit) -> unit
 val or_ : [< sbool ] t -> [< sbool ] t -> [> sbool ] t
+
+(** Similar to [or_], but the rhs is only evaluated if the lhs is not the
+    concrete true. In other words, this is a short-circuiting or. Avoids some
+    errors, like a division by zero in [0 == x || n / x] when [x] is [0]. *)
 val or_lazy : [< sbool ] t -> (unit -> [< sbool ] t) -> [> sbool ] t
+
 val not : [< sbool ] t -> [> sbool ] t
 val distinct : 'a t list -> [> sbool ] t
 val ite : [< sbool ] t -> 'a t -> 'a t -> 'a t


soteria-rust/lib/layout.ml
```diff
@@ -137,34 +138,39 @@ let rec dst_slice_ty : Types.ty -> Types.ty option = function
 let is_dst ty = dst_kind ty <> NoneKind
 
 let size_to_fit ~size ~align =
-  let ( % ) = Stdlib.( mod ) in
-  if size % align = 0 then size else size + align - (size % align)
+  Typed.ite
```
giltho: Do we need `if%val`? :p

soteria-rust/lib/layout.ml
```diff
@@ -189,63 +196,87 @@ let rec layout_of (ty : Types.ty) : layout =
...
   (* Arrays *)
   | TAdt { id = TBuiltin TArray; generics } ->
+      let max_array_len sub_size =
+        (* We calculate the max array size for a 32bit architecture, like Miri does. *)
+        if Typed.equal sub_size Usize.(0s) then
+          (* Special case to avoid a division by zero in the concrete case *)
+          BV.usize Z.(pow (of_int 2) 32)
+        else
+          let isize_bits = 32 - 1 in
+          BV.usize Z.(one lsl isize_bits) /@ Typed.cast sub_size
```
giltho: How is this sound?

giltho (changes requested):
I need a quick answer on that soundness of div by zero before I mark this as approved

Side note: feels good to have time to actually go through PRs

soteria-rust/lib/encoder.ml
```diff
@@ -440,36 +449,36 @@ module Make (Sptr : Sptr.S) = struct
...
                  group the reads together, at least for primitive types. *)
           let arr_ty = mk_array_ty sub_ty len in
-          let layout = layout_of arr_ty in
+          let*** layout = layout_of arr_ty in
           let fields = Seq.init_z len (fun _ -> sub_ty) in
           aux_fields ~f:(fun fs -> Tuple fs) ~layout offset fields
       | TNever -> error `RefToUninhabited
       | TTraitType (tref, name) ->
-          let ty = Layout.resolve_trait_ty tref name in
+          let*** ty = lift_rsymex @@ Layout.resolve_trait_ty tref name in
```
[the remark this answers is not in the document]
N1ark: i mean encoder only has `let***` and `let+++`... which i hate so im changing it now actually ^^

soteria-rust/lib/state.ml
```diff
@@ -265,67 +261,76 @@ let rec check_ptr_align ((ptr, meta) : 'a full_ptr) (ty : Types.ty) st =
...
-let load_discriminant ((ptr, _) as fptr) ty st =
+and load_discriminant ((ptr, _) as fptr) ty st =
   let** (), st = check_ptr_align fptr ty st in
   let parser ~offset = Encoder.variant_of_enum ty ~offset in
   apply_parser ptr parser st
 
+(** Performs a side-effect free ghost read -- this does not modify the state or
+    the tree-borrow state. Returns true if the value was read successfully,
+    false otherwise. *)
+and is_valid_ptr st ptr ty =
```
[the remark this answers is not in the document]
N1ark: 1. good point, will rename 
2. hmmmm maybe? i think it's mostly a thing to ensure lifetimes are correct, i don't think (don't quote me on that) that we catch any additional UB from doing it, since we anyways ensure validity of references etc.

    one thing we _should_ interpret is [`PlaceMention`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.StatementKind.html#variant.PlaceMention), as shown on https://github.com/AeneasVerif/charon/issues/874

soteria-rust/lib/layout.ml
giltho: > note that this isn't a problem when sub_layout.size is symbolic and evaluates to 0 since it makes the condition true anyways, regardless of the / 0.

That was my main issue yeah. As you say, it's prob possible to re-organise code so that it is clear that the div-by-zero case is not unsound because it's not used

N1ark:
ah also forgot to say it here but I ran a before/after benchmark and this seems to give a perf improvement! it says it's ~7% better but it's maybe a bit noisy

giltho:
Wait what how??

PR #204, opened 2025-12-16
Fix frontend hanging with Charon

N1ark:
Small fix for the `ruxt` branch which uses Charon; the frontend would hang on `In_channel.input_lines`, since I believe Charon doesn't print anything to stdout. We now instead only poll for stdout/stderr when requested.

In passing, I also did a small change so that when aliasing checks are disabled, rather than disabling stuff on a per-function basis we just make sure all pointers have no tags. A lot more elegant :)


giltho (approved):
much prettier indeed

Issue #127, opened 2025-09-08
Handle actual Rust layouts


N1ark:
Currently the [layout algorithm](https://github.com/soteria-tools/soteria/blob/main/soteria-rust/lib/layout.ml) (and consequently the [encoding algorithms](https://github.com/soteria-tools/soteria/blob/main/soteria-rust/lib/encoder.ml)) are a weird mix of our own layout calculation using C-based layouts, and extracting *some* information from the frontend (for `enum` niches and tags, notably).

We should stop doing this and instead have a layout that does **no** calculations and just handles parsing the Rust-given layout. This may only work for the [Obol](https://github.com/soteria-tools/obol) frontend, since [Charon](https://github.com/AeneasVerif/charon) doesn't really always do layouts properly I believe (except if using `--monomorphize-experimental`, I guess)

The layout/encoding mechanism should be made into a module with a simple interface, as in the future we will also want to do layouts of polymorphic types. That will require us not relying on the size and align of types being concrete, which is the case in a lot of places.

- [ ] Make layouts symbolic #202 
- [ ] Rely on layouts when encoding (rather than types)
- [ ] Do not calculate layouts except if explicitly required

PR #205, opened 2025-12-18
Actually use layouts 

N1ark:
`Encoder.rust_to_cvals` / `Encoder.rust_of_cvals` where both poorly named functions and badly implemented; in particular they relied very extensively on the type of the values, and handling layouts manually, rather than trusting the layout of the type.

This PR fixes it! I renamed these functions to `encode` and `decode` respectively, and factored out a lot of code and simplified it. The core idea is:
- a value with a `Primitive` layout is written as-is; we don't even check the type
- a value with a `Array` or `Arbitrary` layout has an iterable set fields; `iter_fields` returns an iterator over these fields and their offset
- a value with an `Enum` layout needs special treatment to figure out what variant we're reading/writing; however once we know the variant it's straightforward, as above.

There is relatively little additional effort needed! I also separated the validity checks done for references into its own function, `check_valid` (this comes with the advantage that we can more easily disable validity checks, if we want to one day).

This is the second part of https://github.com/soteria-tools/soteria/issues/127

Additional minor changes:
- added some utilities to `Iter`
- remove the implementation of `Binop.Cmp` for pointers, [as it's never used](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/enum.BinOp.html#variant.Cmp) 
- properly remove metadata from pointers when needed

### DIFF
--- a/soteria-rust/test/cram/cargo.t/run.t
+++ b/soteria-rust/test/cram/cargo.t/run.t
@@ -2,6 +2,6 @@ Test running Rusteria on a crate
   $ soteria-rust cargo . --clean --no-timing
   Compiling... done in <time>
   note: tests::tests::my_test: done in <time>, ran 2 branches
-  PC 1: (0x01 == V|1|) /\ (0x01 == V|1|)
-  PC 2: (0x00 == V|1|) /\ (0x00 == V|1|)
+  PC 1: (V|1| == 0x01) /\ (V|1| == 0x01)
+  PC 2: (V|1| == 0x00) /\ (V|1| == 0x00)
   
--- a/soteria-rust/test/cram/kani.t/run.t
+++ b/soteria-rust/test/cram/kani.t/run.t
@@ -2,11 +2,11 @@ Test kani::any
   $ soteria-rust rustc any.rs --clean --no-timing --kani
   Compiling... done in <time>
   note: any_bool: done in <time>, ran 2 branches
-  PC 1: (0x01 == V|1|) /\ (0x01 == V|1|)
-  PC 2: (0x00 == V|1|) /\ (0x00 == V|1|)
+  PC 1: (V|1| == 0x01) /\ (V|1| == 0x01)
+  PC 2: (V|1| == 0x00) /\ (V|1| == 0x00)
   
   note: any_i8: done in <time>, ran 3 branches
-  PC 1: (0x00 == V|1|) /\ (0x00 == V|1|)
+  PC 1: (V|1| == 0x00) /\ (V|1| == 0x00)
   PC 2: (0x80 <=u V|1|)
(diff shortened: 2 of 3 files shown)


soteria-rust/lib/core.ml
```diff
@@ -10,15 +10,19 @@ module M (Rust_state_m : Rust_state_m.S) = struct
...
-    else if%sat l ==@ r then ok U8.(0s) else ok U8.(1s)
+    if%sat l < r then ok ordering_enum_lt
+    else if%sat l ==@ r then ok ordering_enum_eq else ok ordering_enum_gt
 
   let cmp_of_int v =
     let zero = BV.zero (size_of_int v) in
-    if%sat v <$@ zero then ok U8.(-1s)
-    else if%sat v ==@ zero then ok U8.(0s) else ok U8.(1s)
+    if%sat v <$@ zero then ok ordering_enum_lt
+    else if%sat v ==@ zero then ok ordering_enum_eq else ok ordering_enum_gt
```
giltho: This wouldn't need to branch if we had Rust_val as Rsymex.Value

soteria-rust/lib/state.ml
```diff
@@ -294,17 +293,27 @@ and load_discriminant ((ptr, _) as fptr) ty st =
...
+    | Thin -> true
+    | Len l ->
+        (* TODO: we don't support symbolic slices *)
+        Option.is_some (Typed.BitVec.to_z l)
+    | VTable _ ->
+        (* FIXME: i am not certain how one checks for the validity of a &dyn *)
+        false
+  in
+  if (not can_check_dst) || Option.is_some (Layout.as_zst ty) then
+    return (None, st)
```
giltho: Maybe issue a warning when you can't test some validity?
We need a mechanism for like warn_once.

soteria-rust/lib/layout.ml
```diff
@@ -166,7 +168,10 @@ let rec layout_of (ty : Types.ty) : (t, 'e, 'f) Rustsymex.Result.t =
   | TRawPtr (sub_ty, _)
     when is_dst sub_ty ->
       let ptr_size = Crate.pointer_size () in
-      ok (mk_concrete ~size:(ptr_size * 2) ~align:ptr_size ())
+      ok
+        (mk_concrete ~size:(ptr_size * 2) ~align:ptr_size
+           ~fields:(Array (BV.usizei ptr_size))
+           ())
```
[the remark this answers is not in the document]
N1ark: hmm this could be interesting yes ! i remember @Nadrieril mentioning that slice layouts or their ABI might eventually become guaranteed, so it would be nice to figure out progress on that (pinging because I cannot find a source anywhere:P )

PR #206, opened 2025-12-19
Rust store optimisation

### LINKED ISSUE #162: Store optimisation in Rust
Rust could benefit from the same optimisation soteria-c does, to avoid storing variables into memory if they are in the store, under the assumption they will always be read as a whole, meaning we can skip treeblocks and an expensive encode-decode.

N1ark:
Based on #205 
Fixes #162 

Add the store optimisation to Rust!
Running our Kani + Miri benchmark, we get an average time that goes from 0.232s to 0.211s, so a ~10% improvement! 

Also added a couple reductions, and optimised `bv_to_z` which interprets a bitvector into a signed value; this function went from taking 320 cycles to 31 cycles/op (on one test I tried that was quite slow, where it was used ~4M times)


giltho:
Heh fun idea
@copilot can you promote the files? Check the failed action

soteria-rust/lib/store.ml
```diff
@@ -1,10 +1,50 @@
...
-let find_value sym store = Option.bind (find_opt sym store) fst
+(** We have four kinds of bindings:
+
+    - Stackptr: the symbol is bound to a stack pointer, that lives in the heap.
+    - Value: the symbol is bound to an immediate value; it does not have an
+      address.
+    - Uninit: the symbol is bound to an immediate, uninitialized value.
+    - Dead: the symbol is dead; it doesn't exist (e.g. after a [StorageDead]).
+*)
+type 'a binding_kind =
```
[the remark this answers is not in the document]
giltho: (In a future PR of course)

soteria-rust/lib/store.ml
```diff
@@ -1,10 +1,50 @@
...
+type 'a binding = { kind : 'a binding_kind; ty : Types.ty }
+[@@deriving show { with_path = false }]
+
+type 'a t = 'a binding Map.t
+
+let reserve sym ty =
+  let binding = { kind = Dead; ty } in
+  Map.add sym binding
+
+let[@inline] declare_ sym kind =
```
giltho: Why the underscore?

soteria/lib/bv_values/svalue.ml
```diff
@@ -1794,13 +1795,7 @@ and BitVec : BitVec = struct
...
 
   let of_float ~rounding ~signed ~size v =
-    let p = precision_of_f v.node.ty in
-    match (p, v.node.kind, size) with
-    | F32, Float f, 32 ->
-        mk 32 @@ Z.of_int32 (Int32.bits_of_float (Stdlib.Float.of_string f))
-    | F64, Float f, 64 ->
-        mk 64 @@ Z.of_int64 (Int64.bits_of_float (Stdlib.Float.of_string f))
-    | _, _, _ -> Unop (BvOfFloat (rounding, signed, size), v) <| t_bv size
+    Unop (BvOfFloat (rounding, signed, size), v) <| t_bv size
```
giltho: Curious why this is more efficient?

soteria/lib/soteria_std/list.ml
```diff
@@ -116,3 +116,7 @@ and[@tail_mod_cons] prepend_concat_map2 zs f xs ys =
   match zs with
   | [] -> concat_map2 f xs ys
   | z :: zs -> z :: prepend_concat_map2 zs f xs ys
+
+(** [sub from n l] returns the sublist of [l], from index [from] (inclusive), of
+    length [n]. *)
+let sub from n l = l |> drop from |> take n
```
giltho: Given types of from and n are the same, can we give them label? ~from ~len
Also because it's always ambiguous if it's from-to or from-len

soteria-rust/lib/interp.ml
```diff
@@ -16,88 +16,126 @@ module Make (State : State_intf.S) = struct
   open Rust_state_m
   open Rust_state_m.Syntax
 
-  type store = (full_ptr option * Types.ty) Store.t
-  type 'a t = ('a, store) Rust_state_m.t
+  type 'a t = ('a, Sptr.t Store.t) Rust_state_m.t
   type 'err fun_exec = rust_val list -> (rust_val, unit) Rust_state_m.t
 
-  let get_variable var_id =
+  type lazy_ptr = Store of Expressions.local_id | Heap of full_ptr
```
giltho: How is lazy_ptr not just "place"?
Or possibly "resolved_place" or something

N1ark: well because you don't know if `lazy_ptr` is a lazy pointer; really it's an `(lazy_ptr, full_ptr) Either.t` where `lazy_ptr = local ⊂ place`

soteria-rust/lib/store.ml
N1ark: shrug

soteria-rust/lib/store.ml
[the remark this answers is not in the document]
N1ark: hmmm yes we could!

PR #209, opened 2025-12-22
Share stats printing, improve `Soteria.Terminal`

### LINKED ISSUE #154: Make a shared flag for logging/dumping stats
Currently, soteria-rust and soteria-c both expose a different flag to [log stats to stdout](https://github.com/soteria-tools/soteria/blob/18ed77e88da95d888368644a63a0a3f5bd7764ec/soteria-rust/lib/driver.ml#L100-L116) or [dump them to a file](https://github.com/soteria-tools/soteria/blob/18ed77e88da95d888368644a63a0a3f5bd7764ec/soteria-c/lib/driver.ml#L7-L10), respectively.

We should make shared flags for the stats module, to avoid the duplication.

N1ark:
Fixes #154

A lot more changes than what the issue required but I want the user-facing part of Soteria (i.e. `Terminal`) to be really nice and clean and shareable so the user experience is nice and homogeneous! So I added a bunch of utility that could be shared so we have only one way of output things across the tool.

- Removed `--dump-stats` from Soteria-C and `--stats` from Soteria-Rust; instead added a `Config` module to stats to configure stats outputting, with a flag `--output-stats` (alias `--dump-stats`, `--stats`) to either dump the stats to the given file if specified, or to the terminal if the specified string is `stdout`.
- Added `Terminal.Printers` with some utility printers: `pp_time`, `pp_plural`, `pp_percent`, and `pp_unstable`. `pp_unstable` wraps a printer for "unstable" values, i.e. values that change frequently. If the flag `--hide-unstable` (alias `--diffable`) is passed, `pp_unstable` prints a placeholder value instead (e.g. `<time>`).

  For instance, `pp_time` is wrapped in `pp_unstable`, and in Soteria-Rust I also wrapped code ranges in it, to avoid test diffs when I upgrade Rust versions and the call-traces move around.
- ~~Added `Soteria.Terminal.Config.in_test_environment : unit -> bool`, that returns whether we are currently in a test environment (`dune test`). This is used to automatically enable some flags (`--hide-unstable` and `--clean` in Soteria-Rust), avoiding the need to have them everywhere in the test files.~~

### DIFF
--- a/soteria/lib/terminal/printers.ml
+++ b/soteria/lib/terminal/printers.ml
@@ -5,7 +5,7 @@ let pp_unstable ~name pp ft x =
   if (Config.get ()).hide_unstable then Fmt.pf ft "<%s>" name else pp ft x
 
 (** Pretty-prints a time quantity in seconds, displaying it in either seconds or
-    milli-seconds for smaller values (< 0.05s). Wrapped in [pp_unstable]. *)
+    milli-seconds for smaller values (< 0.05s). Wrapped in {!pp_unstable}. *)
 let pp_time =
   let pp_time ft t =
     if t < 0.05 then Fmt.pf ft "%ams" (Fmt.float_dfrac 2) (t *. 1000.)


soteria-c/lib/driver.ml
```diff
@@ -464,12 +464,14 @@ let generate_all_summaries log_config term_config solver_config config includes
...
-let capture_db log_config term_config solver_config config json_file
-    functions_to_analyse =
+let capture_db log_config term_config solver_config stats_config config
+    json_file functions_to_analyse =
   let open Syntaxes.Result in
   let@ () = with_tool_errors_caught () in
   let functions_to_analyse = as_nonempty_list functions_to_analyse in
-  let@ () = initialise ~log_config ~term_config ~solver_config config in
+  let@ () =
+    initialise ~log_config ~term_config ~solver_config ~stats_config config
```
giltho: We're starting to need a single `~soteria-config` term

soteria-linear/parser/lexer.mll
```diff
@@ -16,11 +16,11 @@ rule read = parse
...
-  | "take" { TAKE }
-  | "requires" { REQUIRES }
-  | "ensures" { ENSURES }
-  | "RW" { RW }
-  | "Freed" { FREED }
+  (* | "take" { TAKE } *)
+  (* | "requires" { REQUIRES } *)
+  (* | "ensures" { ENSURES } *)
+  (* | "RW" { RW } *)
+  (* | "Freed" { FREED } *)
```
giltho: I'm cleaning up some of linear in the logic PR so I'm a bit worried about this creating conflicts, but at the same time I think it's minor enough that it doesn't have to move.

soteria/lib/terminal/printers.ml
```diff
@@ -0,0 +1,23 @@
+(** Wraps a pretty-printer for unstable values, that should not be output into
+    e.g. testing environments where diffs matter. If the config option
+    [hide_unstable] is set, will replace *)
+let pp_unstable ?(name = "x") pp ft x =
+  if (Config.get ()).hide_unstable then Fmt.pf ft "<%s>" name else pp ft x
+
+(** Pretty-prints a time quantity in seconds, displaying it in either seconds or
+    milli-seconds for smaller values (< 0.05s). Wrapped in [pp_unstable]. *)
```
giltho: Don't spend time doing this but fyi you can reference `{!pp_unstable}` to have a link in the doc

soteria/tests/symex/dune
```diff
@@ -1,9 +1,9 @@
 (executables
- (names run_twice deep_give_up if_sure)
+ (names run_twice deep_give_up if_sure print_stats)
```
giltho: Urgh my other PR removes cram tests for symex, it's going to be a mess :(

giltho (approved):
👌

PR #210, opened 2025-12-24
Avoid unnecessary layout calculations

### LINKED ISSUE #127: Handle actual Rust layouts

N1ark:
Fixes #127 
Depends on #205 

- Properly isolate layout computations (`compute_arbitrary_layout`, `compute_enum_layout`) from layout translation from the frontend. The layout is now only computed if the frontend doesn't provide a layout; otherwise it is simply translated as-is. In particular, for the current version of Obol, only closures and tuples require manual calculation.

- Fix the implementation of `intrinsics::size_of_val` and `intrinsics::align_of_val`, taking inspiration of the rustc implementation. The calculation should be a lot more accurate now, which is super important for e.g. the `drop` implementation of `Box<dyn T>`

- More reliably detect uninhabited types

I thought I wanted to do more here but I think this is already better than before and doing more would be over-engineering :) don't fix what isn't broken etc etc etc


giltho (approved):
I'm mostly just trusting this

PR #218, opened 2026-01-08
Pools of solvers

giltho:
Prior to this PR, using `Symex.run` inside an execution (on another process) would be invalid.
It would reuse the same solver state and have banana results.

With this PR, a pool of solver is created and using `Symex.run` will fetch into the pool and get a solver (or create a new one!)
In the future, this will also allow us to run several symbolic execution *in parallel*.

Note: --dump-smt is not thread-safe, this is something to keep in mind for the future!

To implement this:
- I added a new `Resource_pool` module to `Soteria_std`. It provides a thread-safe way of accessing and releasing resource. With additional dynamic checks for use-after-release.
- `Resource_pool` and `Resource_pool_tests` were mostly written by opencode. I have thoroughly read `Resource_pool`, but only read the tests in diagonal. (Though I provided the interface of the pool functor and resource to the LLM, it didn't come up with how to handle this, just the implementation)
- `Reversible` now has a functor to cast a `Mutable` reversible to an "effectful in a pool" reversible which does just the right thing (using `R.run` will fetch a fresh thing in the pool, reset it, and pass it the the function, and the release it at the end).
- I removed all notions of "In place" for solvers 👌

Two additional drive-bys:
- Rewrote `run_needs_stats` as an iterator over the branch that is then given to `Iter.to_list` (which does exactly what we need). We might eventually want to expose the `_iter` versions of functions if users want to interrupt executions in an extrinsic way.
- Rewrote all tests in `soteria/tests/symex` using alcootest instead of dirty cram tests. Much cleaner (done with AI, and then I wrote `interrupt` by hand.

Honest this PR is a testament to beautiful abstractions! The change in `Symex` amounts to using `Solver.Mutable_to_pooled` instead of `Mutable_to_in_place` and everything just clicks the same exact way. And `Solver.Mutable_to_pooled` is just the exact same wrapper as `Solver.Mutable_to_in_place` around `Reversible.Mutable_to_in_place` but using `Reversible.Mutable_to_pooled`. Just beautiful.

See, this just clicks perfectly well into a well-designed framework. So why won't Logic click like that? 😭


N1ark (approved):
veryyyy nice could you check perf on Collections-C very quickly? i doubt it will matter at all but just in case

giltho:
Before this PR: 
```
Benchmark 1: dune exec -- time soteria-c capture-db /Users/sacha/Prog/soteria-c-experiments/Collections-C/build/compile_commands.json --solver-timeout 2 --dump-stats /Users/sacha/Prog/soteria-c-experiments/results/Collections-C/stats.json --use-cerb-headers --havoc-undef
  Time (mean ± σ):      2.057 s ±  0.019 s    [User: 0.737 s, System: 0.276 s]
  Range (min … max):    2.031 s …  2.091 s    10 runs
```

After this PR:
```
Benchmark 1: dune exec -- time soteria-c capture-db /Users/sacha/Prog/soteria-c-experiments/Collections-C/build/compile_commands.json --solver-timeout 2 --dump-stats /Users/sacha/Prog/soteria-c-experiments/results/Collections-C/stats.json --use-cerb-headers --havoc-undef
  Time (mean ± σ):      2.060 s ±  0.009 s    [User: 0.744 s, System: 0.270 s]
  Range (min … max):    2.050 s …  2.080 s    10 runs
```

So I'd say the change is only noise

PR #225, opened 2026-01-16
Bugfix: do not block on child process stdout

CharlyCst:
Previously, `soteria-rust` was waiting first on stdout, and then on stderr when starting a child process. This causes problem when the stderr buffer gets full, as writes to stderr become blocking and because soteria-rust is waiting on stdout to return End Of File before draining stderr we get a deadlock.

This commit fixes the issue by reading from both stdout and stderr at the same time, using the usual `select` syscall in a loop until both stream reach End Of File. Unfortunately we have to do things using low-level APIs as the higher-level OCaml abstractions tend to be blocking.

### DIFF
--- a/soteria-rust/lib/frontend.ml
+++ b/soteria-rust/lib/frontend.ml
@@ -34,7 +34,47 @@ module Exe = struct
 
   let is_ok = function Unix.WEXITED 0 -> true | _ -> false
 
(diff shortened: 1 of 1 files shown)


N1ark (approved):
Looks great, thank you very much :)

PR #234, opened 2026-01-21
Add benchmarks to CI

giltho:
This is in progress. Currently, I need to figure out a lean way of exporting Soteria-Rust as well as Soteria-C in order to run in a separate runner.


giltho:
Next step: since #265 has been merged, we can use the packages to run the benchmarks in parallel processes!

PR #244, opened 2026-01-28
Revamp `Logs`/`Terminal` split

N1ark:
- Redefine the CLI of `Logs` the same way we do everything else
- Move `Terminal.Color,Profile,Printers` to `Logs.Color,Profile,Printers`
- Move the `pp_clr` etc. printers from `Color` to `Printers`
- Colour the severity of logs in stdout
- Made all levels 5 characters long, adding space as padding if necessary (not shown in images)

Screenshots in a few different Zed themes (since themes change terminal colors):

image image image


N1ark:
!build-doc

PR #251, opened 2026-01-30
Extract Symex.Core from Symex.Base, reduce code in State_monad

### LINKED ISSUE #224: Improve `State_monad`
Right now, we have [State_monad](https://github.com/soteria-tools/soteria/blob/main/soteria/lib/sym_states/state_monad.ml). It's heavy, and it's a lot of code duplication. That code can probably be simplified.

giltho:
In an attempt to de-duplicate the code between symex.ml and state_monad.ml, I extracted a subset of `Symex.Base` into `Symex.Core`, with a functor that produces "the rest".

This is going to be much better since in #217 I drastically extend the signature of `Symex.Base`, which leads to a huge amount of code duplication. Now, it should be prevented

Fixes #224

### DIFF
--- a/soteria/lib/symex/symex.ml
+++ b/soteria/lib/symex/symex.ml
@@ -31,6 +31,7 @@ module type Core = sig
   (** Represents a yet-to-be-executed symbolic process which terminates with a
       value of type ['a]. *)
   type 'a t
+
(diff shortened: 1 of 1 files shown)


giltho:
!build-doc

N1ark (approved):
nice

PR #252, opened 2026-01-30
Script to synchronise versions

giltho:
This PR proposes a script to update versions:
```
usage: versionsync.py [-h] {list,check,update,set} ...

Version synchronization script for Soteria

positional arguments:
  {list,check,update,set}
    list                List all configured versions
    check               Check if all versions are in sync
    update              Update all version occurrences
    set                 Set a version and update all occurrences

options:
  -h, --help            show this help message and exit

Examples:
  versionsync.py list                    List all configured versions
  versionsync.py check                   Check if all versions are in sync
  versionsync.py update                  Update all version occurrences
  versionsync.py set OCAML_VERSION 5.5.0 Set a version and update everywhere

Filters:
  Tags can include filters to transform values:
  [versionsync: NAME=value] [versionsync.slice: 0..3]
  
  Slice examples:
    0..3   - first 3 characters
    0..-1  - all but last character
    ..-2   - all but last 2 characters
    2..    - from index 2 to end
  ```
  
The best way I found to avoid having over-complicated regexps is to have all versions twice, once in a tag and once in the actual version. When seeing a tag, we then do a simple search/replace for the first occurence of that exact text in the next 5 lines (unless another tag is encountered before 5 lines).
If the value is not found (i.e. if it was modified by hand by someone and desynchronised from the tag), it also returns an error.

There's a simple way of filtering as well so that one can write
```
# [versionsync: OCAML_VERSION=5.6] [versionsync.slice: ..-2]
ocaml-version = 5.6
```

I decided to do that instead of X.X(.X) so that it can be any string. It's slightly easier to maintain...

### CODE REVIEW

**Comment by giltho** [comment]
image

image

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -20,10 +20,13 @@ concurrency:
   cancel-in-progress: true
 
 env:
+  # [versionsync: OCAML_VERSION]
   OCAML_VERSION: 5.4.0
   OBOL_REPO: soteria-tools/obol
   OBOL_COMMIT_HASH: 45f2f95cf034c5daa448b64945e721839a57bb22
```
N1ark: can you add a config for `OBOL_COMMIT_HASH`?

scripts/versionsync.py
```diff
@@ -0,0 +1,252 @@
...
+# replacement_template: string with {version} placeholder
+VERSION_LOCATIONS: list[tuple[str, str, str, str, bool]] = [
+    # ci.yml - OCAML_VERSION
+    (
+        ".github/workflows/ci.yml",
+        "OCAML_VERSION",
+        r"(# \[versionsync: OCAML_VERSION\]\n  OCAML_VERSION: )[^\n]+",
+        r"\g<1>{version}",
+        False,
+    ),
```
N1ark: .... interesting choice of format but i suppose it works so I won't complain

N1ark (changes requested):
Add obol version, and commands to list versions and edit a version through the commandline please
The comment on the formatting being weird is a nitpick, if you want to bother doing it in a more consistent way it's cool and will make the script actually maintainable but i dont mind too much

scripts/versionsync.py
```diff
@@ -0,0 +1,252 @@
+#!/usr/bin/env python3
+"""
+Version synchronization script for Soteria.
+
+This script manages version strings across the codebase to ensure consistency.
+Versions are defined in scripts/versions.json and propagated to various files.
+
+Usage:
+    scripts/versionsync.py check   - Check if all versions are in sync
+    scripts/versionsync.py update  - Update all version occurrences
```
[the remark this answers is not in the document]
N1ark: i prefer having to only access one thing so only the script or only the file

giltho:
Updated PR description

N1ark (approved):
niceeeee

PR #259, opened 2026-02-03
Generate flamegraphs during Symex

giltho:
I'm not sure this is ready to be merged. You'll see the changes are rather minimal.
There's an outstanding bug that I need to figure out, but so far it seems like it's holding up pretty good :) 

It literally it Just Works™ like this in the interpeter. This is the *only* changed made in the Soteria-C interpreter:
image
And in the Rust interpreter:
image

The only issue is the lifting of the various run functions... I wonder if we should somehow provide those with `State_monad`

And you'll see that the changes in Symex are actually rather minimal as well! Just calling `push` and `pop` frame in the right places, and some additional `Flamegraph.checkpoint` just in case.

### DIFF
--- a/soteria-rust/lib/interp.ml
+++ b/soteria-rust/lib/interp.ml
@@ -1169,6 +1169,7 @@ module Make (StateImpl : State.S) = struct
       with_extra_call_trace ~name:fundef.item_meta.name
         ~loc:fundef.item_meta.span.data ~msg:"Entry point"
     in
+    let@ () = StateM.with_frame "Entry point" in
     let generics = TypesUtils.generic_args_of_params () fundef.generics in
     let* value = exec_real_fun fundef generics args in
     let* () = State.run_thread_exits () in
--- a/soteria-rust/lib/interp.ml
+++ b/soteria-rust/lib/interp.ml
@@ -1133,8 +1133,8 @@ module Make (StateImpl : State.S) = struct
       | None -> Fmt.kstr not_impl "Function %a is opaque" Crate.pp_name name
       | Some body -> ok body
     in
-    let string_name = Fmt.to_to_string Crate.pp_name name in
-    let@ () = Rust_state_m.with_frame string_name in
+    let name_str = Fmt.to_to_string Crate.pp_name name in
+    let@ () = Rust_state_m.with_frame name_str in
     Soteria.Stats.As_ctx.incr StatKeys.function_calls;
     let@@ () = Poly.push_generics ~params:fundef.generics ~args:generics in
     let@@ () = with_env ~env:Store.empty in
--- a/soteria-rust/lib/interp.ml
+++ b/soteria-rust/lib/interp.ml
@@ -1134,7 +1134,7 @@ module Make (StateImpl : State.S) = struct
       | Some body -> ok body
     in
     let name_str = Fmt.to_to_string Crate.pp_name name in
-    let@ () = Rust_state_m.with_frame name_str in
+    let@ () = StateM.with_frame name_str in
     Soteria.Stats.As_ctx.incr StatKeys.function_calls;
     let@@ () = Poly.push_generics ~params:fundef.generics ~args:generics in
     let@@ () = with_env ~env:Store.empty in
--- a/soteria-rust/lib/interp.ml
+++ b/soteria-rust/lib/interp.ml
@@ -1133,6 +1133,8 @@ module Make (StateImpl : State.S) = struct
       | None -> Fmt.kstr not_impl "Function %a is opaque" Crate.pp_name name
       | Some body -> ok body
     in
+    let string_name = Fmt.to_to_string Crate.pp_name name in
+    let@ () = Rust_state_m.with_frame string_name in
     Soteria.Stats.As_ctx.incr StatKeys.function_calls;
     let@@ () = Poly.push_generics ~params:fundef.generics ~args:generics in
     let@@ () = with_env ~env:Store.empty in
@@ -1167,6 +1169,11 @@ module Make (StateImpl : State.S) = struct
       with_extra_call_trace ~name:fundef.item_meta.name
         ~loc:fundef.item_meta.span.data ~msg:"Entry point"
     in
+    (* FIXME: If using this, this throws an error.
+       The abstraction is leaking, execution of immediate things is done *before* execution starts. *)
+    (* Fmt.pr "REACHED HERE@.@?";
+    let@ () = Rust_state_m.with_frame "Entry point" in
+    Fmt.pr "REACHED HERE 2@.@?"; *)
     let generics = TypesUtils.generic_args_of_params () fundef.generics in
     let* value = exec_real_fun fundef generics args in
     let* () = State.run_thread_exits () in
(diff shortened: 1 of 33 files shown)


soteria-rust/lib/interp.ml
```diff
@@ -1126,12 +1126,14 @@ module Make (State : State_intf.S) = struct
       | None -> Fmt.kstr not_impl "Function %a is opaque" Crate.pp_name name
       | Some body -> ok body
     in
+    let string_name = Fmt.to_to_string Crate.pp_name name in
+    let@ () = Rust_state_m.with_frame string_name in
```
N1ark:
```suggestion
    let name_str = Fmt.to_to_string Crate.pp_name name in
    let@ () = Rust_state_m.with_frame name_str in
```

soteria-rust/lib/rustsymex.ml
```diff
@@ -41,10 +41,10 @@ end
 include Soteria.Sym_states.State_monad.Make (MonoSymex) (MonadState)
 include Syntaxes.FunctionWrap
 
-let run_with_stats ?fuel ?fail_fast ~mode symex =
+let run_with_stats ?flamegraph_svg ?fuel ?fail_fast ~mode symex =
   run_with_state ~state:MonadState.empty symex
   |> (Fun.flip MonoSymex.map) fst
-  |> MonoSymex.Result.run_with_stats ?fuel ?fail_fast ~mode
+  |> MonoSymex.Result.run_with_stats ?flamegraph_svg ?fuel ?fail_fast ~mode
```
N1ark: sidenote i dont really like `flamegraph_svg` as a name...... `flamegraph` would be nicer? even if it's with some type `flamegraph_out = | File of string | None | Stdout`

btw do we want to do stdout printing in this PR or at a later date?

soteria/lib/profiling/flamegraph.ml
```diff
@@ -0,0 +1,97 @@
...
+    let flamegraph = ref Flamegraph.empty in
+    let current_stack : Flamegraph.stack Dynarray.t = Dynarray.create () in
+    let () = Dynarray.add_last current_stack (Flamegraph.stack []) in
+    let last_checkpoint = ref (Unix.gettimeofday ()) in
+    let map_stack (f : Flamegraph.stack -> Flamegraph.stack) =
+      let last = Dynarray.pop_last current_stack in
+      Dynarray.add_last current_stack (f last)
+    in
+    let checkpoint () =
+      let current_time = Unix.gettimeofday () in
```
N1ark: not for now but something fun could be to add a C component to your flamegraphs library to return the number of instructions (like in landmarks) and then also keep track of that here

soteria/lib/profiling/flamegraph.ml
```diff
@@ -0,0 +1,97 @@
...
+              Dynarray.truncate current_stack (len - n);
+              continue k ()
+          | effect Save, k ->
+              Dynarray.add_last current_stack (Dynarray.get_last current_stack);
+              continue k ()
+          | effect Checkpoint, k ->
+              checkpoint ();
+              continue k ())
+    in
+    res
```
N1ark:
```suggestion
    Fun.protect
      ~finally:(fun () ->
        checkpoint ();
        match Flamegraphs.Svg.to_file flamegraph_svg !flamegraph with
        | Ok () -> ()
        | Error e ->
            Logs.L.error (fun m -> m "Could not write flamegraph: %s" e))
      (fun () ->
        try f () with
        | effect Map_stack f, k ->
            checkpoint ();
            map_stack f;
            continue k ()
        | effect Backtrack_n n, k ->
            checkpoint ();
            let len = Dynarray.length current_stack in
            Dynarray.truncate current_stack (len - n);
            continue k ()
        | effect Save, k ->
            Dynarray.add_last current_stack (Dynarray.get_last current_stack);
            continue k ()
        | effect Checkpoint, k ->
            checkpoint ();
            continue k ())
```

N1ark (changes requested):
a lot of nitpicks but id rather we get this right because i'm already slightly unconvinced about how much this is complexifying `Symex`; like it's not a lot but having to add checkpoints in these subtle places seems like an easy thing to get wrong (maybe I'm mistaken).
either way this is missing a tiny bit of documentation. i'd have liked having stdout output because i dont really like using external apps to view output if possible, but that can be done later since i imagine it needs changes in Flamegraph.
also what's the bug you mention in the PR description?

soteria/lib/symex/symex.ml
```diff
@@ -633,7 +653,9 @@ module Make_core (Sol : Solver.Mutable_incremental) = struct
         (with_section @@ fun () -> a () f);
         loop r
 
-  let vanish () _f = ()
+  let vanish () _f =
+    Flamegraph.checkpoint ();
+    ()
```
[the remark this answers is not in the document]
giltho: For what it's worth, this is exactly how I think of Soteria's Symex. It can be as arbitrarily complex as necessary to provide all the Very Nice Stuff, I don't care, that's the point.
But yeah, as I said, I'm not even sure that one is necessary :p

soteria-rust/lib/interp.ml
```diff
@@ -1163,6 +1165,11 @@ module Make (State : State_intf.S) = struct
     let@@ () =
       with_extra_call_trace ~loc:fundef.item_meta.span.data ~msg:"Entry point"
     in
+    (* FIXME: If using this, this throws an error.
+       The abstraction is leaking, execution of immediate things is done *before* execution starts. *)
+    (* Fmt.pr "REACHED HERE@.@?";
+    let@ () = Rust_state_m.with_frame "Entry point" in
+    Fmt.pr "REACHED HERE 2@.@?"; *)
```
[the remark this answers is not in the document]
giltho: Fixed

soteria/lib/soteria_std/unix_soteria.ml
```diff
@@ -0,0 +1,18 @@
+include Unix
+
+let dir_exists path =
```
N1ark:
```suggestion
(** [dir_exists path] returns true if [path] exists and is a directory, false if
    [path] does not exist, and raises an error if [path] exists but is not a
    directory. *)
let dir_exists path =
```

soteria/lib/soteria_std/unix_soteria.ml
```diff
@@ -0,0 +1,18 @@
...
+  match Sys.is_directory path with
+  | true -> true
+  | false -> Fmt.failwith "Expected a directory but found a file: %s" path
+  | exception Sys_error _ -> false
+
+let ensure_dir_exists path =
```
N1ark, line 16:
```suggestion
(** [ensure_dir_exists path] ensures that [path] exists and is a directory,
    creating it and any missing parent directories if necessary. Raises an error
    if [path] exists but is not a directory. *)
let ensure_dir_exists path =
```

N1ark (approved):
lgtm, thank you! i think we are approaching the point where we should figure out a way of plugging things in `Symex` to avoid cluttering it with additional things; it's ok for now but if we e.g. end up with Callgraph + Flamegraph + Stats + Coverage etc it might quickly become a lot (though maybe im overthinking it)

unrelated but having a `profiling` module and a `stats` module kind of seems odd, to me they are similar things. merge as is but maybe eventually we should merge stats and flamegraph into profiling or something of the sort (e.g. dev_tools...? idk)

soteria/lib/soteria_std/unix_soteria.ml
giltho: Now that I read this doc, this function makes no sense lmao
If the thing is not a dir it exn??? If the thing is nothing it just says false???

PR #262, opened 2026-02-05
Add `versionsync pull` command

giltho:
```
./versionsync.py pull obol
./versionsync.py pull charon
```

respectively go to the directory ../charon and ../obol (which can be customised with --dir), fetches the correct hash from the correct repo, and builds the new version.

```
./versionsync.py pull all
```
does both


N1ark (approved):
thanks!

Issue #126, opened 2025-09-08
☂️ Documentation


giltho:
This is the umbrella issue for the documentation of Soteria. The list is:

- [ ] Document every module
  - [ ] bv_values
  - [ ] c_values
  - [ ] logs
  - [ ] solvers
  - [ ] sym_data
  - [ ] sym_states
  - [ ] symex
  - [ ] terminal
  - [ ] stats

Also, we need to make a decision of how much documentation we want before we release 0.1

giltho:
So I'd rather avoid doing other things that require my time this week unless I'm confident I can give it to the slop machine. Possibly  documentation of each module is easy... Tutorials not sure

PR #266, opened 2026-02-08
Document `solvers` and some of `symex`

N1ark:
Document `Soteria.Solvers`, and part of `Soteria.Symex`, as part of closing #126 
Although the initial doc was LLM generated, I went through all of it and review it / shortened it where relevant.

### DIFF
--- a/soteria/lib/solvers/value.ml
+++ b/soteria/lib/solvers/value.ml
@@ -1,6 +1,10 @@
 open Simple_smt
 
-(** The type of values the solver operators on. *)
+(** The type of values the solver operators on.
+
+    This is different from {!Symex.Value.S}, which represents symbolic values in
+    the context of symbolic execution. Here we are simply interested in how some
+    arbitrary typed values can be encoded into SMT-LIB. *)
 module type S = sig
   type t
   type ty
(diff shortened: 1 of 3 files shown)


N1ark:
aaa CI broken bc of Z3 \~.~

soteria/lib/solvers/value.ml
```diff
@@ -0,0 +1,17 @@
+open Simple_smt
+
+(** The type of values the solver operators on. *)
```
giltho: Same, why is this different from `Soteria.Symex.Value.S`

soteria/lib/symex/compo_res.ml
```diff
@@ -1,3 +1,25 @@
+(** This module defines a three-way result type for compositional symbolic
```
giltho: Note that `miss` can be generalised to any incompleteness of the engine actually. `miss` really just means "the analysis couldn't conclude but *maybe* there's a way to add anti-frame to fix it. Not there *must* be some

soteria/lib/symex/solver_result.ml
```diff
@@ -8,6 +12,15 @@ let pp fmt = function
...
 
+(** [admissible ~mode result] determines whether a branch should be taken based
+    on the solver result and approximation mode.
+
+    The decision matrix:
+    - [Sat, _]: Always take the branch (definitely feasible)
+    - [Unsat, _]: Never take the branch (definitely infeasible)
+    - [Unknown, OX]: Take the branch (over-approximate: don't miss paths)
+    - [Unknown, UX]: Don't take the branch (under-approximate: avoid possibly
+      unsatisfiable paths) *)
```
giltho: Possibly an actual matrix like, and explanation below? No strong opinion, just a proposition
```ocaml
(**
Decision matrix:
    {table
      {tr {td  } {th SAT } {th UNSAT } {th UNKNOWN } }
      {tr {th UX } {td true } {td false } {td {b false} } }
      {tr {th OX } {td true } {td false } {td {b true} } }
    }
*)
```

giltho (approved):
I wrote this review a while ago and realised I never submitted it...

PR #269, opened 2026-02-09
fix merging of statistics in experiments script

giltho:
Script was broken since the redo fo statistics


giltho:
There's a bit of a bug in the handling of RawJson in Soteria.Stats btw.

Ranges are serialized arrays `[ [ sline, scol ], [ eline, ecol] ]`
(I.e. a pair of pairs)
Because they're rawjson, when you have a list of them, instead of getting a list of pairs of pairs, you get just a list of pairs, they're concatenated.

PR #272, opened 2026-02-11
Fuzzer for simplifying constructors

giltho:
I think I'm happy with the state of the code as-is.
I'm happy with the code as-is. I've basically entirely rewritten the generator, which was the messy part.

The only issue is that we can't merge it because... it's still finding bugs.

- The fuzzer creates expressions and simplifies them, and then checks for equality through z3. Soundness should require that all tests pass, even on corner cases.
- On merge queue, and only when the `bv_solver` folder was modified, the fuzzer is run with 40k tests (8 x 5k)
- On demand, through `!run-bv-fuzzer` or by dispatching the workflow on main, the fuzzer is run with 120k tests (8 x 15k)
- Added a `?force` flag to `eval`, which re-evaluates every node even if children haven't changed.
- Fixed a bunch of unsoundnesses in the simplifications the code found by the fuzzer

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,17 +118,23 @@ jobs:
         if: github.event_name == 'merge_group'
         id: check-bv-files
         run: |
-          # Get the base branch commit (the commit before merge queue)
-          BASE_SHA=$(git merge-base origin/main HEAD)
+          BASE_SHA="${{ github.event.merge_group.base_sha }}"
+          HEAD_SHA="${{ github.event.merge_group.head_sha }}"
+          echo "Comparing $BASE_SHA..$HEAD_SHA"
           
-          # Check if any files in soteria/lib/bv_values were modified
-          if git diff --name-only $BASE_SHA HEAD | grep -q "^soteria/lib/bv_values/"; then
+          # Use the GitHub API to compare commits (avoids needing full git history)
+          CHANGED=$(gh api repos/${{ github.repository }}/compare/${BASE_SHA}...${HEAD_SHA} \
+            --jq '.files[].filename' | grep -c "^soteria/lib/bv_values/" || true)
+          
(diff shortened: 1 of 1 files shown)


soteria/lib/bv_values/svalue.ml
```diff
@@ -1438,21 +1441,22 @@ and BitVec : BitVec = struct
         zero (size_of v1.node.ty)
     | Binop (Shl, v, { node = { kind = BitVec s1; _ }; _ }), BitVec s2 ->
         let n = size_of v1.node.ty in
-        shl v (mk n Z.(s1 + s2))
+        let total = Z.(s1 + s2) in
+        if Z.(total >= of_int n) then zero n else shl v (mk n total)
```
N1ark: same here there is already a reduction for this so you can just trust `shl`

soteria/lib/bv_values/svalue.ml
N1ark: and here too id rather you just remove the `total >= of_int n` branch and add a proper reduction where if the rhs is concrete and `>= n`, it becomes `ashr lhs (mk n (n - 1))`

soteria/lib/bv_values/svalue.ml
```diff
@@ -1992,9 +2009,14 @@ and BitVec : BitVec = struct
   let mul_overflows ~signed v1 v2 =
     match (v1.node.kind, v2.node.kind) with
     | BitVec l, BitVec r -> ovf_check ~signed (size_of v1.node.ty) l r Z.( * )
+    | _ when signed && size_of v1.node.ty == 1 ->
+        (* We need to special-case size one because the other simplifications
+           will mes with that case, and it's pretty easy to simplify. *)
```
```suggestion
        (* We need to special-case size one because the other simplifications
           will mess with that case, and it's pretty easy to simplify. *)
```

PR #276, opened 2026-02-22
Improvements on Soteria on user experience for Soteria-C scripts

giltho:


soteria/lib/symex/symex.ml
```diff
@@ -364,6 +364,12 @@ module StatKeys = struct
   *)
   let miss_without_fix = "soteria.miss-without-fix"
 
+  (** Number of times [branch_on] was called *)
+  let branch_on_calls = "soteria.branch-on-calls"
+
+  (** Number of times [branch_on] actually branched *)
+  let branch_on_branched = "soteria.branch-on-branched"
```
N1ark: these can stay, however, i think

soteria-c/lib/csymex.ml
```diff
@@ -1,5 +1,14 @@
 module SYMEX = Soteria.Symex.Make (Bv_solver.Z3_solver)
 
+module Concrete_alloc_id = struct
+  let next_id = ref 1
+
+  let get_next_id () =
+    let id = !next_id in
+    incr next_id;
+    id
```
N1ark: what about something like this?
```suggestion
  let get_next () =
    let id = !next_id in
    incr next_id;
    Typed.Ptr.loc_of_int id
```

PR #281, opened 2026-03-02
Cleanup Validity in Rust

N1ark:
- Rework the way we check for validity; rather than having it a bit scattered around (when reading Rtree_block leaves, when decoding a value, when creating a nondet value), there is one central function called `validity`:
  ```ocaml
  val validity : 
	?check_ref:(full_ptr -> ty -> (unit, 'e, 'f) Result.t) ->
    ty ->
    rust_val ->
    (([> Typed.sbool ] Typed.t * Error.t) list, 'e, 'f) Result.
  ```
  This function returns all the validity constraints on a value; users can then e.g. `assert_or_error` all checks with the associated error, or `assume` all constraints to ensure the value is valid.
- Remove `Value_codec.nondet`, in favour of `Value_coded.nondet_raw` to create a "raw" nondet value (without validity constraints checked) and `nondet_valid` for `nondet_raw |> assume validity`.
- Implement the required SL constructs for `Tree_state` & co: `produce`, `serialize` etc. For now we are extremely minimal and only produce the heap, nothing else. That will come later! This just unblocks some other future work.
- Add support for a hypothetical compositional mode in Soteria Rust

~~**TODO:** I need to re-run the Miri benchmark to check for regressions around validity checks and ensure nothing got lost. I will also take it as an opportunity to run a quick perf check; I think technically what we do now is _slower_, my hopes is just that it's not noticeable.~~ This seems to have <1% impact

### DIFF
--- a/soteria-rust/bin/soteria_rust.ml
+++ b/soteria-rust/bin/soteria_rust.ml
@@ -61,6 +61,4 @@ end
 let cmd =
   Cmd.group (Cmd.info ~exits "soteria-rust") [ Exec.cmd; Build_plugins.cmd ]
 
-let () =
-  Printexc.record_backtrace true;
-  exit @@ Cmd.eval cmd
+let () = exit @@ Cmd.eval cmd


soteria-rust/bin/soteria_rust.ml
```diff
@@ -61,4 +61,6 @@ end
 let cmd =
   Cmd.group (Cmd.info ~exits "soteria-rust") [ Exec.cmd; Build_plugins.cmd ]
 
-let () = exit @@ Cmd.eval cmd
+let () =
+  Printexc.record_backtrace true;
```
giltho: Maybe only activate this in some level of verbose? It makes execution more expensive

PR #299, opened 2026-03-25
Add `[@@deriving sym_state]`

N1ark:
I'd recommend reviewing + merging #309 first, so the diff here is much cleaner

The much requested PPX is here!

```ocaml
type t = { heap : Heap.t option; globs : Globs.t option }
[@@deriving sym_state { symex = Csymex }]
```

that's it :) [tutorial](https://soteria-tools.github.io/soteria/api/pr-299/soteria/sym_state_ppx.html)

generates a state model that satisfies `Sym_states.Base.M (Csymex).S`. This means:
- creating the state monad `SM`
- generating `syn`, the `to_syn` and `ins_outs` functions
- generating `produce` and `consume`
it also defines two utility wrappers for all fields:
```ocaml
val with_heap :
  ('a, 'e, Heap.serialized list) Heap.SM.Result.t ->
  ('a, 'e, serialized list) SM.Result.t

val with_heap_sym : 'a Heap.SM.t -> 'a SM.t

(* and then with_globs, with_globs_sym *)
```

furthermore, it supports:
- **ignored fields**:  fields that don't represent a state model, with `[@sym_states.ignore { empty }]`, where `empty` is the empty value to use.
- **lifting between state monads**: allows tweaking the behaviour of the `with_field` wrapper to lift to a specific state monad. for instance in Soteria Rust, with `DecayMapMonad`, we do:
  ```ocaml
  pointers : DecayMap.t option;
  heap : Heap.t option; [@sym_state.context { field = pointers }]
  ```
- **`syn` equality**: if you need to assert that `syn` is equal to some external type (e.g. in soteria-c we have `type syn = State_intf.syn = ...`, you can do so by annotating the deriving:
  ```ocaml
  type t = { heap : Heap.t option; globs : Globs.t option }
  [@@deriving sym_state { symex = Csymex; syn = State_intf.syn }]
  ```

In a future PR i will try adding sum type support, which would enable us to formulate Freeable as `Freed of Freed_ex.t option | Init of S.t option` !!

the ppx is ~600 lines, that i tried to make as readable as possible after the slopper generated it. you can check out the results in soteria-c and soteria-rust's states : they are a bit simpler ! or rather, they are more complex : the boring boilerplate is gone, so only the stuff that requires work is there :) 

i wouldnt review commit by commit, as they overlap significantly


N1ark:
!build-doc

soteria/tutorial/sym_state_ppx.mld
```diff
@@ -0,0 +1,160 @@
...
+val consume : syn -> t option -> (t option, syn list) My_symex.Consumer.t
+
+val with_heap :
+  ('a, 'e, Heap.syn list) Heap.SM.Result.t ->
+  ('a, 'e, syn list) SM.Result.t
+
+val with_heap_sym : 'a Heap.SM.t -> 'a SM.t
+]}
+
+[with_heap_sym] is a wrapper that calls a symbolic computation of type [Heap.SM.t] with the heap part of the state and updates it with the result. On the other hand, [with_heap] is a more powerful wrapper that also lifts missing outcomes into the right [syn] variant.
```
giltho, line 67: It took me a full minute to understand this sentence.
I'm being annoying but could we instead have `with_heap_res` and `with_heap`? I still find the `_sym` suffix confusing

giltho (approved):
At this point I'm wondering if it's worth adding a `Or_zero ≃ Option` module, and force every field to be `Or_zero.t`.

soteria/tutorial/sym_state_ppx.mld
N1ark: the thing is `with_heap_res` is what people are most likely to use, so it feels "wrong" having it have the longer name. if you think it's ok then i dont mind ill change it

N1ark:
> At this point I'm wondering if it's worth adding a `Or_zero ≃ Option` module, and force every field to be `Or_zero.t`.

please no more modules 😭

giltho:
- please no more modules 😭
- OCaml
Pick 1

soteria/tutorial/sym_state_ppx.mld
giltho: Hmm, unless we can find a better suffix than sym it itches

N1ark: _base ?

giltho: Meh

giltho: Ok let's keep with_heap_sym for now

N1ark:
> * please no more modules 😭
> * OCaml
>   Pick 1

theyre just quite heavyweight and we already have a lot of them (adding your https://github.com/soteria-tools/ppx_mixins would make it easier)

N1ark:
+ having something that basically means `Option` makes learning the library less intuitive imo; we should try using OCaml things when they fit

Issue #301, opened 2026-03-26
Move borrow state/protector reasoning into `Tree_borrow`


N1ark:
Currently the interface of `Tree_borrow` expects the client to specify e.g. what state to start reborrows in. This is fine for now, but we could do something [like in rustc](https://github.com/rust-lang/rust/blob/3a933e5bf5b695faf9e09d47a263a20c6b85c26a/src/tools/miri/src/borrow_tracker/mod.rs#L265) where information is abstracted away; we only need to know the borrow kind, and the borrowed type.

this comes with two advantages:
- `TreeState` becomes entirely unaware of anything relating to tree borrows, which is good (it should never have been the case)
- this would also allow us to support stacked borrows super easily, which although is probably not of much interest is a nice to-have, and would probably require very little effort to implement

PR #308, opened 2026-04-06
Parametric Tree Borrows

N1ark:
First step towards allowing more ways of doing alias-checks. This makes the whole state parametric on a "Tree_borrows" module, which also supports reasoning with logic if desired! The interface might adapt as we discover gaps or additional needs, but I think this is a relatively solid start.

- Functorise `Tree_state`, which now receives a `Symex -> Tree_borrows_intf.S` functor. It's a functor to allow different state implementations that may not use `DecayMapMonad` to use it
- (Finally) move the implementation of `Sptr` into `State_intf`. There is still a `sptr.ml` file that defines a decay map and the "small" Sptr interface for `Decoder`, but it still makes things simpler as we don't have to lie about the pointer being independent from the state.
- Cleaned up the `Sptr` interface to remove extra functions we don't want/need.

Functorising Tree_state means the PR diff is massive and unreadable; I've organised the commits so the ugly diff change happens only in f0a88f2, which is only indentation; every other commit should have a clean diff.

This will enable #301


soteria-rust/lib/builtins/intrinsics_impl.ml
```diff
@@ -144,7 +145,7 @@ module M (StateM : State.StateM.S) : Intrinsics_intf.M(StateM).Impl = struct
   let caller_location : full_ptr ret =
     (* TODO: we should really do something better here *)
     let+ () = ok () in
-    (Sptr.null_ptr (), Thin)
+    (Sptr.of_address Usize.(0s), Thin)
```
giltho: I don't like this, can we keep some `null_ptr` helper around? I don't like obtaining "null_ptr" by casting 0 everywhere

N1ark:
@giltho no rush in reviewing this, but fyi ive reorganised the commits so it should be fairly easy to review per-commit

soteria/lib/sym_states/tree_block.ml
```diff
@@ -101,7 +101,8 @@ module MemVal (Symex : Symex.Base) = struct
         state can be composed with it; in other words, calling [produce] on a
         tree with this node must always vanish. Otherwise this should raise a
         [miss] with the fixes needed for this to become exclusively owned. *)
-    val assert_exclusively_owned : t -> (unit, 'err, syn list) Symex.Result.t
+    val assert_exclusively_owned :
+      (t, sint) tree -> (unit, 'err, syn list) Symex.Result.t
```
giltho, line 105: Why ?

soteria/lib/sym_states/tree_block.ml
```diff
@@ -146,11 +147,12 @@ module Make (Symex : Symex.Base) (MemVal : MemVal(Symex).S) = struct
...
     let is_fully_owned = function NotOwned _ -> false | Owned _ -> true
 
-    let assert_exclusively_owned = function
+    let assert_exclusively_owned t =
+      match t.node with
       | NotOwned _ ->
           Result.miss_no_fix
             ~reason:"assert_exclusively_owned - tree not fully owned" ()
-      | Owned n -> MemVal.assert_exclusively_owned n
+      | Owned _ -> MemVal.assert_exclusively_owned t
```
giltho: Seems redundant if you're passing the whole thing to `MemVal.assert_exclusively_owned` anyway?
```ocaml
let assert_exclusively_owned = MemVal.assert_exclusively_owned
```

soteria-rust/lib/state/tree_state.ml
```diff
@@ -6,11 +6,120 @@ module T = Typed.T
...
+  let fresh () = failwith "Fresh unimplemented for sptr (for now)"
+
+  let subst subst_val p =
+    let se = Typed.Expr.subst subst_val in
+    let ptr = se p.ptr in
+    let align = se p.align in
+    let size = se p.size in
+    { ptr; align; size; tag = p.tag }
+
+  let null_ptr () =
```
giltho: Wait you actually have this, so why are you using `of_address 0` in other places?

soteria/lib/logic/util.ml
```diff
@@ -0,0 +1,27 @@
+module M (Symex : Symex.Base) = struct
+  (** Utilities to handle common types with Soteria's logic modules. *)
+
+  module Abstr = Data.Abstr.M (Symex)
+
+  module Option = struct
+    (** Option type lifted to the value world. Implements
+        {{!Soteria.Abstr.M.S_with_syn}[S_with_syn]}. *)
```
giltho, line 8: That's not true if you don't implement fresh?

soteria/lib/logic/util.ml
```diff
@@ -0,0 +1,27 @@
...
+
+    type 'a t = 'a option [@@deriving show { with_path = false }]
+    type 'a syn = 'a option [@@deriving show { with_path = false }]
+
+    (* TODO: do we have [let fresh () = return None] ? *)
```
giltho: It'd need to be something like
```ocaml
let fresh ~fresh_inner () =
	branch [
		fun () -> return None;
	    fun () -> let+ inner = fresh_inner () in Some inner
	]
```
Option should be an abstraction combinator (and can be derived with a PPX probably)

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -135,93 +114,266 @@ module Make (Sptr : Sptr.S) = struct
...
+      | SZeros, Owned (Leaf (Init _, _)) ->
+          lift @@ not_impl "Assume rust_val == 0s"
+      | SZeros, Owned (Leaf _) -> lfail Typed.v_false
+      (* tree borrows *)
+      | STree_borrow_st s, Owned (Leaf (v, tb)) ->
+          let+ tb' =
+            let+? fixes = Borrows.State.consume s tb in
+            List.map lift_tb_st_fix fixes
+          in
+          mk_leaf t v tb'
```
giltho: I don't fully understand the implications of this for now.
And we need to tell Pedro what to do not to hit bugs in there if we merge it

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -135,93 +114,266 @@ module Make (Sptr : Sptr.S) = struct
...
-       SZeros, Owned (Zeros, _) -> ok (not_owned t) | SZeros, Owned (Init _, _)
-       -> not_impl "Assume rust_val == 0s" | SZeros, _ -> vanish () *)
+    let consume (s : syn) (t : tree) : (tree, syn list) DecayMap.SM.Consumer.t =
+      let open DecayMap.SM.Consumer in
+      let open Syntax in
+      match (s, t.node) with
+      | _, NotOwned _ -> miss_no_fix ~reason:"rtree_block consume notowned" ()
+      | _, Owned Lazy -> lift @@ not_impl "Consume on lazy node"
+      (* init *)
+      | SInit _, _ ->
```
giltho: `Rust_val.learn_eq` not doing the trick?

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -135,93 +114,266 @@ module Make (Sptr : Sptr.S) = struct
...
-       (s, t.node) with | _, NotOwned _ -> miss [] (* init *) | SInit _, _ ->
-       not_impl "Consume typed value on rust_val equality." (* any *) | SAny,
-       Owned _ -> ok (not_owned t) (* uninit *) | SUninit, Owned (Uninit
-       Totally, _) -> ok (not_owned t) | SUninit, _ -> vanish () (* zeros *) |
-       SZeros, Owned (Zeros, _) -> ok (not_owned t) | SZeros, Owned (Init _, _)
-       -> not_impl "Assume rust_val == 0s" | SZeros, _ -> vanish () *)
+    let consume (s : syn) (t : tree) : (tree, syn list) DecayMap.SM.Consumer.t =
+      let open DecayMap.SM.Consumer in
+      let open Syntax in
+      match (s, t.node) with
```
giltho: Nit: rewrite this to be
```ocaml
let++ (mv', tb') = match (s, t.node) with (...) in 
mk_leaf t mv' tb'
```
Ensures you don't accidentally return some (Unowned, tb) when you modify this out of context in the future?

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -135,93 +114,266 @@ module Make (Sptr : Sptr.S) = struct
...
+                let* v = Producer.apply_subst (Rust_val.subst Sptr.subst) v in
+                let _, middle = l.range in
+                let+^ vl, vr = split_rval v middle in
+                (* HACK: is this sound? Doing it this way because we can't have
+                   a split_rval for exprs, given it requires decaying pointers.
+                   An alternative would be to have a [produce_leaf] which takes
+                   in a [Leaf _], so we only subst once. *)
+                let vl = Rust_val.to_syn Sptr.to_syn vl in
+                let vr = Rust_val.to_syn Sptr.to_syn vr in
+                (SInit vl, SInit vr)
```
giltho, line 283: Not the most elegant but somewhat correct.

giltho: Ideally, if you have a solution to subst only once, I'd prefer that. But don't spend too much time right now, we're bound to test this code a *lot* more in the near future, it will evolve anyway

soteria-rust/lib/state/tree_state.ml
```diff
@@ -23,47 +28,46 @@ module Make () = struct
     }
```
giltho, line 28: Honestly, `[@@deriving syn]`?

soteria-rust/lib/state/tree_state.ml
```diff
@@ -210,24 +213,48 @@ module Make () = struct
     open SM
     open Syntax
 
+    let lift_fix_block (s : Tree_block.syn) =
+      (* HACK: this is quite ugly; tree borrow misses relating to the structure
+         (not the state!) can occur within [Tree_block], however fixing them
+         there is not possible, since the structure is stored here, in [Block].
+         As such we catch those, and lift them appropriately here. *)
```
giltho, line 220: Yeah, it feels like the tb/state/syn story isn't finalised yet. 
Honestly it's ok, as long as it doesn't block Pedro, or as long as you're ready to provide support whenever he's blocked

giltho (approved):
Mostly comments, questions, and minor nits

soteria/lib/sym_states/tree_block.ml
N1ark: because with the change to `R_tree_block`, you may have `Owned` trees that aren't exclusively owned and for which you need to recurse on the tree, e.g. `Lazy -> Leaf (NotOwned, Some tb_state), Leaf (Zeros, None)`

soteria-rust/lib/state/tree_state.ml
N1ark: i think you're reviewing an intermediary commit; this function is removed later on

soteria/lib/logic/util.ml
N1ark: oh sure that works, idk why i hadnt considered branching
meh i dont mind hand-implementing the base types for now, what i care about is just having them exist so we can easily derive the other ones

soteria-rust/lib/state/tree_state.ml
N1ark: yes that's why i made that utils module, i unfortunately dont want to bother implementing and im consistently out of tokens to have the slopper do it :/ but yes that's the plan

soteria-rust/lib/state/tree_state.ml
N1ark: i dont understand; as far as Pedro is concerned he just uses the `Concrete` tree borrows implementation so none of this will happen anyways

N1ark: i agree there is still work to do for this btw; but im unclear how to handle it cleverly so for now im just doing what works :P

soteria-rust/lib/state/tree_state.ml
N1ark: bikeshedding but i think the syntax is going to have to be something like this:
`[@@deriving syn { symex = MySymex; values = Typed.t }]`
because "symex-values" like Typed don't have to expose `learn_eq` & co.

soteria-rust/lib/state/rtree_block.ml
N1ark: yeah ill leave this as is unfortunately simply bc of time (id really want to get this merged because it's conflict-prone)

soteria/lib/sym_states/tree_block.ml
giltho: Hmm, shouldn't one just always recurse entirely? I'm not sure it would be much more expensive

soteria-rust/lib/state/tree_state.ml
giltho: But then can we bring it back? 😂

soteria/lib/logic/util.ml
giltho: That being said, I dream of a world where calling fresh on something like this would dynamically declare a new ADT kind to the solver so that we can actually have nondet of this adt

soteria-rust/lib/state/tree_state.ml
N1ark: yes yes im addressing your reviews

soteria-rust/lib/state/tree_state.ml
giltho: We can expose it in the outer layer for the PPX to work, that's pretty easy, but yeah we'll figure out the needs as we write it

soteria-rust/lib/state/tree_state.ml
giltho: >  dont understand; as far as Pedro is concerned he just uses the Concrete tree borrows implementation so none of this will happen anyways

Ah yes, fair

soteria-rust/lib/state/rtree_block.ml
N1ark: hadnt realised it exists ^^

soteria-rust/lib/state/rtree_block.ml
N1ark: done; i actually just handle v and tb entirely separately which makes it really cute

soteria/lib/sym_states/tree_block.ml
N1ark: yeah possibly, thats another option

PR #309, opened 2026-04-07
Small state improvements

N1ark:
Improvements to state monads done while working on #299 
This is best reviewed commit-by-commit !

### DIFF
--- a/soteria-c/lib/state_intf.ml
+++ b/soteria-c/lib/state_intf.ml
@@ -1,6 +1,5 @@
 open Typed
 open T
-open Cerb_frontend
 module Agv = Aggregate_val
 module Compo_res = Soteria.Symex.Compo_res
 
@@ -38,7 +37,7 @@ module type S = sig
 
   val produce_aggregate :
     Expr.t ->
-    Ctype.ctype ->
+    Cerb_frontend.Ctype.ctype ->
     Aggregate_val.syn ->
     t option ->
     t option Csymex.Producer.t
--- a/soteria-rust/lib/sptr.ml
+++ b/soteria-rust/lib/sptr.ml
@@ -15,10 +15,9 @@ open T
     [from_exposed] is partial: if a provenance cannot be guessed from the
     integer, [None] is returned. *)
 module type DecayMapS = sig
-  type t
+  include Soteria.Sym_states.Base.M(Rustsymex).S
 
   val empty : t
-  val pp : Format.formatter -> t -> unit
 
   (** Decays the given location into an integer, updating the decay map
       accordingly. If [expose] is true, the provenance is marked as exposed, and
@@ -29,19 +28,16 @@ module type DecayMapS = sig
     size:[< sint ] Typed.t ->
     align:[< nonzero ] Typed.t ->
     [< sloc ] Typed.t ->
-    t ->
-    (sint Typed.t * t) Rustsymex.t
+    sint Typed.t SM.t
 
   (** Tries finding, for the given integer, the matching provenance in the decay
       map. If found, it returns that provenance, along with the exposed address
       for that allocation at offset 0. Otherwise returns [None]. *)
   val from_exposed :
-    [< sint ] Typed.t ->
-    t ->
-    ((sloc Typed.t * sint Typed.t) option * t) Rustsymex.t
+    [< sint ] Typed.t -> (sloc Typed.t * sint Typed.t) option SM.t
 end
 
-module DecayMap : DecayMapS = struct
+module DecayMap = struct
   module MapKey = struct
     include Typed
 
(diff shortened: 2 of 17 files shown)


soteria-c/lib/state_intf.ml
```diff
@@ -38,7 +37,7 @@ module type S = sig
 
   val produce_aggregate :
     Expr.t ->
-    Ctype.ctype ->
+    Cerb_frontend.Ctype.ctype ->
     Aggregate_val.syn ->
     t option ->
     t option Csymex.Producer.t
```
giltho, line 43: ?

giltho: I mean it's not worse but is there a reason?

soteria-rust/lib/sptr.ml
```diff
@@ -52,81 +48,105 @@ module DecayMap : DecayMapS = struct
...
+
+    let exprs_syn ({ address; exposed = _ } : syn) : Expr.t list = [ address ]
+
+    let subst s ({ address; exposed } : syn) : t =
+      { address = Expr.subst s address; exposed }
   end
 
-  type entry = { address : sint Typed.t; exposed : bool }
-  [@@deriving show { with_path = false }]
+  module EntryExcl = Soteria.Sym_states.Excl.Make (Rustsymex) (Entry)
```
giltho: Agreement?

soteria/lib/sym_states/excl.ml
```diff
@@ -22,10 +20,9 @@ module Make (Symex : Symex.Base) (E : Data.Abstr.M(Symex).S_with_syn) = struct
   open SM
   open SM.Syntax
 
-  let assert_exclusively_owned st : (unit, 'err, syn list) Symex.Result.t =
-    let open Symex.Syntax in
-    let+ res, _ = unwrap () st in
-    Compo_res.map res (fun _ -> ())
```
giltho, line 28: heh

soteria/lib/sym_states/freeable.ml
```diff
@@ -13,8 +13,11 @@ module Make
     (I : sig
       include Base.M(Symex).S
 
-      val assert_exclusively_owned :
-        t option -> (unit, 'err, syn list) Symex.Result.t
+      (** Checks this state is exclusively owned, returning if it is exclusively
+          owned and otherwise missing with the fixes required to make this state
+          exclusively owned.
+          {b This function is expected to not be modify the state.} *)
```
giltho, line 19: I'm being a bit pedantic here, but maybe this should really just be:
`t option -> (unit, 'err, syn list) Result.t`

giltho: (Though not it's fine to modify the symbolic state to something equivalent, e.g. to rebalance TreeBlocks)

soteria/lib/symex/symex.ml
```diff
@@ -281,6 +281,16 @@ module type Base = sig
     val run : subst:subst -> 'a t -> ('a * subst) symex
     val run_identity : 'a t -> 'a symex
 
+    (** Exposes the underlying type of this producer monad. This is unsafe and
+        shouldn't be used in clients, it is only available to enable the
+        implementation of the state monad transformer. *)
+    val leak_UNSAFE : 'a t -> subst option -> ('a * subst option) symex
```
giltho: This doesn't need to be exposed.
```ocaml
let leak_NOT_REALLY_UNSAFE_ACTUALLY_IF_YOU_THINK_ABOUT_IT m s_opt =
  match s_opt with
   | None ->
      let+ res = Producer.run_identity m in
     (res, None)
   | Some subst ->
       let+ (res, subst) = Producer.run ~subst m in
       (res, Some subst)
```
You can just implement it in SM and not expose it. I agree it's not pretty, but this direction of abstraction doesn't need any unsafe

soteria/lib/symex/symex.ml
```diff
@@ -281,6 +281,16 @@ module type Base = sig
...
+
+    (** The dual of [leak_UNSAFE] for producers. This is unsafe and shouldn't be
+        used in clients, it is only available to enable the implementation of
+        the state monad transformer. *)
+    val unleak_UNSAFE : (subst option -> ('a * subst option) symex) -> 'a t
```
giltho: `from_raw_UNSAFE`? `unleak` is a bit weird

This is the direction that's dangerous. The Producer and Consumer operations are carefully crafted so that you can only use the subst in a way that is expected. By using any `subst -> subst` function there's a danger.

soteria/lib/sym_states/state_monad.ml
```diff
@@ -115,6 +128,30 @@ module Make
...
   end
 
+  module Producer = struct
+    include Producer
+
+    let[@inline] run_with_state ~state (x : 'a t) : ('a * st) Symex.Producer.t =
+      let open Symex.Syntax in
+      Symex.Producer.unleak_UNSAFE (fun subst ->
+          let+ (res, subst), state = leak_UNSAFE x subst state in
+          ((res, state), subst))
```
giltho, line 147: See the `leak_UNSAFE` comment from earlier.
Apart from this, I was nerdsnipped for a while trying to figure out a solution, didn't find one. 
We could create an opaque `opaque_subst` type, and then expose `Producer.t = opaque_subst -> ('a * opaque_subst) Symex.t`. And then run takes a 'a Producer.t and returns a proper `subst`. This would make it safe, as it would ensure we can't construct a producer from bananas.
Now... the huge capital letters `leak_UNSAFE` should be safeguard enough

giltho (approved):
Just change the leak thing

soteria-c/lib/state_intf.ml
N1ark: yeah i think it shadows a type i need layer ? im not sure actually i think i don't need this anymore

soteria-rust/lib/sptr.ml
N1ark: ah .. yeah true oops

PR #312, opened 2026-04-09
Bump Charon and Obol

N1ark:
Test changes are:
- mostly that function names are now prefixed with the crate name
- the `assert!` macros are somehow not overriden by Soteria anymore? this would require further investigation but it's minor, it mostly impacts quality of error messages

I also had to split frontend into a frontend_runtime and frontend so I could use information about the toolchain in error diagnostics without creating circular dependencies.

Currently running the benchmarks to check for regressions


N1ark:
Converted to draft; major regressions

demo/poly-vec.rs
```diff
@@ -0,0 +1,31 @@
...
+
+//
+
+//
+
+//
+
+//
+
+//
```
giltho: //

demo/sll.rs
```diff
@@ -1,235 +1,65 @@
...
-pub struct LinkedList<T> {
-    head: Option<NonNull<Node<T>>>,
-    tail: Option<NonNull<Node<T>>>,
-    len: usize,
-    marker: PhantomData<Box<Node<T>>>,
-}
+// Kani: kani demo/sll.rs
+// Soteria Rust: dune exec -- soteria-rust exec ./demo/sll.rs --kani --compact
 
-struct Node<T> {
```
giltho, line 11: Maybe remove the balle perdue pour Kani, and rename the file to dll, because it's def not an sll

giltho:
REUSE file needs to be updated to account for files moving

giltho (approved):

demo/sll.rs
N1ark: ill just remove this i never use it

PR #317, opened 2026-04-17
Use ppx_mixins

giltho:
What are we thinking about this?

Also, I'm realising that I don't quite know why S_with_syn is not just a `with_syn` module without the `S`

### DIFF
--- a/soteria.opam
+++ b/soteria.opam
@@ -61,5 +61,5 @@ build: [
 dev-repo: "git+https://github.com/soteria-tools/soteria.git"
 pin-depends: [
   ["simple_smt.~dev" "git+https://github.com/NatKarmios/simple-smt-ocaml#9db7ef0bb61a76010de255af3cbe4066dd8dd622"]
-  ["ppx_mixins.~dev" "git+https://github.com/soteria-tools/ppx_mixins#041919eb78dc643c88e6e570a896fdd9c41e25f6"]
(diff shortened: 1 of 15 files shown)


soteria.opam
```diff
@@ -60,4 +61,5 @@ build: [
 dev-repo: "git+https://github.com/soteria-tools/soteria.git"
 pin-depends: [
   ["simple_smt.~dev" "git+https://github.com/NatKarmios/simple-smt-ocaml#9db7ef0bb61a76010de255af3cbe4066dd8dd622"]
+  ["ppx_mixins.~dev" "git+https://github.com/soteria-tools/ppx_mixins#041919eb78dc643c88e6e570a896fdd9c41e25f6"]
```
N1ark: did you try publishing to opam?

N1ark (approved):
yes i think that's nicer :) i wonder if maybe using `+` or `and` rather than a `;` in the PPX would be nicer, as it'd mean ocamlformat won't try inserting line returns for every module
```ocaml
type before
[@@mixins
    D_abstr.S_with_syn;
    D_abstr.Sem_eq]

type after
[@@mixins D_abstr.S_with_syn + D_abstr.Sem_eq]
```

soteria.opam
giltho: Not yet, I'm not going to publish something we don't agree on

giltho:
Quite a good idea :)

giltho:
If we merge this I'm so going to abuse that pattern for abstractions all over the place ❤️

giltho:
https://github.com/ocaml/opam-repository/pull/29741

PR #319, opened 2026-04-19
Use released version of ppx_mixins

giltho:
Also looks like the latest release of `linol` broke the API, so I fixed it to `0.10`


giltho:
Merging this without approval, there cannot be debate and I don't want to base my next branch on something that has a weird pin

N1ark (approved):

PR #323, opened 2026-04-23
A few fixes on how learning of variables work

giltho:
- The previous behaviour when learning new variables could be created was wrong (reporting a tool bug)
- If `e` is already fully known, we don't have to do anything.

- Also, I asked Claude to produce all learnable things it was convinced of. It found a few nice ones, though I had to correct the code it was all over the place

@pcarrott can you please read through Expr.learn and tell me if you thing anything that you believe could be wrong?
After that, you can rebase your work on this PR and see if everything goes thorugh

### DIFF
--- a/soteria/lib/bv_values/expr.ml
+++ b/soteria/lib/bv_values/expr.ml
@@ -61,24 +61,66 @@ module Subst = struct
         let vs, s = apply_list ~missing_var s vs in
         (v :: vs, s)
 
-  let learn (s : t) (e : Svalue.t) (v : Svalue.t) : t option =
-    let rec aux s (e : Svalue.t) v =
-      match e.node.kind with
-      | Bool _ | Float _ | BitVec _ -> Some s
-      | Var _ -> if Raw_map.mem e s then Some s else Some (extend e v s)
-      | Unop (Unop.Not, e') -> aux s e' (Bool.not v)
-      (* This pattern matching can be extended for any reversible operation. *)
-      | _ -> None
-    in
+  let is_known (s : t) (e : Svalue.t) : bool =
     let exception Not_covered in
     try
-      (* If the substitution already covers the value, just return the input
-         substitution *)
       let _ = apply ~missing_var:(fun _ _ -> raise_notrace Not_covered) s e in
-      Some s
-    with Not_covered ->
-      (* Otherwise, we try learning. *)
-      aux s e v
+      true
+    with Not_covered -> false
+
+  let rec learn (s : t) (e : Svalue.t) (v : Svalue.t) : t option =
+    let open Syntaxes.Option in
+    let/ () = if is_known s e then Some s else None in
+    match e.node.kind with
+    | Var _ -> if Raw_map.mem e s then Some s else Some (extend e v s)
+    (* Boolean negation: Not(e') = v => e' = Not(v) *)
+    | Unop (Unop.Not, e') -> learn s e' (Bool.not v)
+    (* Bitwise NOT is self-inverse: BvNot(e') = v => e' = BvNot(v) *)
+    | Unop (Unop.BvNot, e') -> learn s e' (BitVec.not v)
+    (* Arithmetic negation is self-inverse mod 2^n: Neg(e') = v => e' =
+       Neg(v) *)
+    | Unop (Unop.Neg, e') -> learn s e' (BitVec.neg v)
+    (* Bit extension: BvExtend(false, by)(e') = v => e' = extract(lower bits of
+       v). This is valid whether the extension is signed or unsigned. *)
+    | Unop (Unop.BvExtend (_, _by), e') ->
+        let size = size_of e'.node.ty in
+        learn s e' (BitVec.extract 0 (size - 1) v)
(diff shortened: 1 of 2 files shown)


soteria/lib/bv_values/expr.ml
```diff
@@ -61,11 +61,65 @@ module Subst = struct
...
+    (* Bitwise NOT is self-inverse: BvNot(e') = v => e' = BvNot(v) *)
+    | Unop (Unop.BvNot, e') -> learn s e' (BitVec.not v)
+    (* Arithmetic negation is self-inverse mod 2^n: Neg(e') = v => e' =
+       Neg(v) *)
+    | Unop (Unop.Neg, e') -> learn s e' (BitVec.neg v)
+    (* Bit extension: BvExtend(false, by)(e') = v => e' = extract(lower bits of
+       v). This is valid whether the extension is signed or unsigned. *)
+    | Unop (Unop.BvExtend (_, _by), e') ->
+        let size = size_of e'.node.ty in
+        learn s e' (BitVec.extract 0 (size - 1) v)
```
N1ark, line 87: i know this is correct but my way of thinking about it is wrong and i dont know why

my intuition says this is wrong because
`extend(unsigned, by 2, 0b00) = 0b1100` is false and `0b00 = 0b00` is true,
but ofc here this is "correct" i just dont understand why theoretically; is this an implication? i.e. the way to think about these is 
`(extend(unsigned, by 2, 0b00) = 0b1100) ==>  0b00 = 0b00`

soteria/lib/bv_values/expr.ml
```diff
@@ -61,11 +61,65 @@ module Subst = struct
...
+        match v.node.kind with
+        | BitVec z when Z.equal z Z.zero -> learn s e' Bool.v_false
+        | BitVec z when Z.equal z Z.one -> learn s e' Bool.v_true
+        | _ -> None)
+    (* Addition: e1 + c = v => e1 = v - c (or symmetrically) *)
+    | Binop (Binop.Add _, e1, e2) ->
+        if is_known s e1 then learn s e2 (BitVec.sub v e1)
+        else if is_known s e2 then learn s e1 (BitVec.sub v e2)
+        else None
+    (* Subtraction: e1 - c = v => e1 = v + c c - e2 = v => e2 = c - v *)
```
N1ark:
```suggestion
    (* Subtraction: e1 - c = v => e1 = v + c, c - e2 = v => e2 = c - v *)
```

soteria/lib/symex/symex.ml
```diff
@@ -1040,23 +1040,21 @@ module Base_extension (Core : Core) = struct
...
-            tool_bug
-              "Tool Bug: learned substitution does not cover expression's free \
-               variables.")
-          subst expr
-      in
-      assert_pure (Value.sem_eq_untyped v v') subst
+      match Value.Expr.Subst.learn subst expr v with
+      | None ->
+          L.debug (fun m ->
+              m "Couldn't learn %a := %a" Value.Expr.pp expr Value.ppa v);
```
N1ark: (it's going to take us a bit to get used to it huh..)
```suggestion
          [%l.debug
              "Couldn't learn %a := %a" Value.Expr.pp expr Value.ppa v];
```

N1ark (approved):

soteria/lib/symex/symex.ml
giltho: Heh

pcarrott:
Thanks, Sacha! This seems to make sense and consumption is behaving as expected. I'm now getting a bug on producers tho: `Stdlib.Effect.Unhandled(Soteria__Soteria_std__Reversible.Mutable_to_pooled(M).Wrap(_))` is being raised when I try to produce the summary `V|1| - V|1| % 2`. When calling `Symex.Producer.apply_subst`, the error appears due to the call of `Solver.fresh_var` on type `(TBitVector 32)`, which is called by `nondet_UNSAFE`. Since in, this case, `subst` has no knowledge of `V|1|`, I think the idea is to create a fresh variable here, but for some reason it is crashing while doing so.

N1ark:
@pcarrott are you using the producer from within a `Symex.run` call?

soteria/lib/bv_values/expr.ml
giltho: So, first of all, let's not that there is no soundness requirement on this function, the goal is to *try* learning some substitution. If the learned substitution is wrong, it will get caught later.

So we're trying to *match* against an assertion. In particular, imagine your *current* (linear) heap is a single binding `0 -> 0b1100` where `v` is some concrete bitvector.
And now, let's say the assertion you're trying to match against is `0 -> extend(unsigned, 2, x)`.
You need to learn a substitution for `x`. The *only* possible case is if `x` is the last two bytes of `0b1100`.

After that, `learn_eq` will still assert `0b0000 = 0b1100` and fail, but this is still the correct substitution to learn (because if we had `0 -> 0b0000` in our state instead, it'd be the correct one).

giltho:
Let's take this to Zulip. Merging this in the meantime

PR #326, opened 2026-04-27
Clean up tech debt

N1ark:
Bunch of small things to make the codebase maybe nicer; each commit is one issue we had open.
Some of the changes are not as cute as I was expecting (the first two, notably), so we can always decide to only apply some of these changes and write down on the issue what didn't work on this attempt for the next attempt at fixing them.

- Swap arguments of all `map`/`bind` to be `map f x` rather than `map x f`, enabling nicer currying. Took it as an opportunity to make sure `Iter`, `Seq`, `Option`, `Result` and `List` satisfy `Monad.S` inside `Soteria_std`
- Make `S_int` not depend on a `S_bool`, and make it slightly more sensible. Until now it was a weird sort of thing where it was meant to be used as `S_int.t Symex.Value.t`, which is relatively impractical and doesn't compose too well. `S_int` is now it's own standalone type. This caused a bit of churn for `Tree_block` and `Plist` but i think it's better
- Moved `Compo_res` from `Soteria.Symex` to `Soteria.Soteria_std`, which is usually open. This is a net positive :) 
- Expand `Monad` and `Compo_res` to provide more module types and functors; this then allows trimming quite a lot of boilerplace from `Symex`, and means all our monads expose all the folders `fold_list`, `fold_iter`, iter_list`, etc. etc. etc. with minimal repetition. This is an increase in LoC but I think it's worth it


giltho (approved):
Very good!

PR #327, opened 2026-04-29
Better symex run

N1ark:
Commits are self-contained:
- cleaned up `Flamegraph` to be more like `Stats`
- add a `Effects.Managed_effect` module type for modules that raise effects that can be managed in different ways (notably: ignored, dumped or handled externally),
- simplify the signature of `Symex.run` and merge all variants (`_needs_stats`, `_with_stats`) exploiting that
- Tried sharing more code between `Symex.run` and `Symex.Result.run`; not super convinced about it but eh

### DIFF
--- a/soteria/lib/profiling/flamegraph.ml
+++ b/soteria/lib/profiling/flamegraph.ml
@@ -48,12 +48,15 @@ module Make (M : Monad.Base) = struct
   let checkpoint () = Effect.perform Checkpoint
 
   let with_frame (name : string) (f : unit -> 'a M.t) : 'a M.t =
-    push_frame name;
-    M.map
-      (fun r ->
-        pop_frame ();
-        r)
-      (f ())
+    (* the bind + return to make sure the effect isn't raised until the
+       computation is ran. *)
+    M.return ()
+    |> M.bind @@ fun () ->
+       push_frame name;
+       f ()
+       |> M.map @@ fun r ->
+          pop_frame ();
+          r
 
   let with_ _name f =
     let stacks = ref [] in
--- a/soteria/lib/profiling/flamegraph.ml
+++ b/soteria/lib/profiling/flamegraph.ml
@@ -1,5 +1,4 @@
 open Soteria_std
-open Syntaxes.FunctionWrap
 
 type frame = string
 type stack = { rev_frames : frame list; weight : float }
@@ -11,7 +10,7 @@ let pop_stack (stack : stack) =
   let rev_frames = List.tl stack.rev_frames in
   { stack with rev_frames }
 
-let write_folded_stacks path stacks =
+let dump path stacks =
   let path =
     if String.ends_with ~suffix:".collapsed" path then path
     else path ^ ".collapsed"
@@ -38,6 +37,8 @@ module Make (M : Monad.Base) = struct
     | Save : unit Effect.t
     | Checkpoint : unit Effect.t
 
+  type t = stack list
+
   let push_frame (s : string) : unit =
     Effect.perform (Map_stack (push_string_to_stack s))
 
@@ -54,7 +55,7 @@ module Make (M : Monad.Base) = struct
         r)
       (f ())
 
-  let with_dumped flamegraph f =
+  let with_ _name f =
     let stacks = ref [] in
     let current_stack : stack Dynarray.t = Dynarray.create () in
     Dynarray.add_last current_stack
@@ -72,12 +73,11 @@ module Make (M : Monad.Base) = struct
       stacks := { stack with weight = elapsed } :: !stacks
     in
     let open Effect.Deep in
-    let@ () =
-      Fun.protect ~finally:(fun () ->
-          checkpoint ();
-          write_folded_stacks flamegraph (List.rev !stacks))
-    in
-    try f () with
+    try
+      let res = f () in
+      checkpoint ();
+      (res, List.rev !stacks)
+    with
     | effect Map_stack g, k ->
         checkpoint ();
         map_stack g;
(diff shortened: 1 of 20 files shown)


giltho:
About to go to sleep but how about instead of `Managed_effect` we call it `Bookkeeping` or `Effectful_bookkeeping`?

That would justify the "dumped" thing, and it makes the purpose of this interface clearer?

N1ark:
> About to go to sleep but how about instead of `Managed_effect` we call it `Bookkeeping` or `Effectful_bookkeeping`?
> 
> That would justify the "dumped" thing, and it makes the purpose of this interface clearer?

sure!

soteria/lib/symex/symex.ml
```diff
@@ -30,6 +30,18 @@ let () =
...
+type 'a effect_handling =
+  | Ignore
+      (** Handle the effect by ignoring it, so don't collect any information. *)
+  | Dump of 'a
+      (** Handle the effect, and dump the information according to the relevant
+          configuration. The value is the parameter passed to the feature's dump
+          handler. *)
+  | Handled
+      (** The effect is already handled by the caller, nothing needs to be done.
+      *)
```
giltho: Nitpicking but "Handled" doesn't work. "Handled" by who? I initially as "handled by the function"

Bit pendantic but let's called this `Handed_by_caller`?

soteria/lib/profiling/flamegraph.ml
```diff
@@ -45,21 +47,20 @@ module Make () = struct
...
+    (* the bind + return to make sure the effect isn't raised until the
+       computation is ran. *)
+    M.bind (M.return ()) @@ fun () ->
+    push_frame name;
+    M.map (f ()) (fun r ->
+        pop_frame ();
+        r)
 
-  let run ~flamegraph f =
+  let with_ _name f =
```
giltho, line 61: Something went wrong with the root name?

soteria/lib/symex/symex.ml
```diff
@@ -424,24 +429,8 @@ module type S = sig
...
-      ?flamegraph:string ->
-      ?fuel:Fuel_gauge.t ->
-      ?fail_fast:bool ->
-      mode:Approx.t ->
-      ('ok, 'err, 'fix) t ->
-      (('ok, 'err Or_gave_up.t, 'fix) Compo_res.t * Value.(sbool t) list) list
-      Stats.with_stats
-
-    val run_needs_stats :
-      ?flamegraph:string ->
```
giltho, line 389: feels good

giltho (approved):
Absolutely

soteria/lib/symex/symex.ml
N1ark: i meant it in the se de of "it's handled"; your option sounds a bit long... then how about just Caller or External/Extern ?

giltho: > i meant it in the se de of "it's handled"

Yes yes, I understand, but it's confusing
Hmmm, `Caller` it is

soteria/lib/profiling/flamegraph.ml
N1ark: if you're talking about the `_name`,  that's something i don't know how to make prettier : the arg argument goes in both with_ and with_dumped, but actually in this case we only need it in with_dumped, so here its ignored

giltho: Right but then there's something weird somewhere else let me put a comment there

soteria/lib/symex/symex.ml
```diff
@@ -377,40 +389,33 @@ module type S = sig
...
+      the execution is performed and what information is collected during
+      execution:
+      - [fuel] receives a {{!Fuel_gauge.t}fuel gauge} to limit execution depth
+        and breadth. By default, the fuel is infinite.
+      - [stats] specifies whether statistics about the execution are already
+        handled outside the function (see {!Stats.As_ctx.with_}) or should be
+        handled and ignored by this function (default).
+      - [flamegraph] specifies whether a flamegraph should be created from this
+        symbolic process or if it should be ignored (default). The value is the
+        name of the top-level frame in the flamegraph.
```
giltho: Here it is: what does "the value is the name of the top-level frame in the flamegraph?"

N1ark: how about this then
```diff
       - [flamegraph] specifies whether a flamegraph should be created from this
-        symbolic process or if it should be ignored (default). The value is the
-        name of the top-level frame in the flamegraph.
+        symbolic process or if it should be ignored (default). The value passed
+        in the [Dump] variant is the name of the top-level frame in the
+        flamegraph.
```

giltho:
I don't understanddddddd
You have:
- The value passed in the dumped variant is the value that is given to `with_dumped` right?
- In `Flamegraph.with_dumped`, this same value is passed to `with_`, and it's entirely ignored. 

The documentation says that this value will be the name of the top-level frame in the graph but won't it always be ` "SYMBOLIC EXECUTION ROOT"`?

N1ark:
> I don't understanddddddd You have:
> 
> * The value passed in the dumped variant is the value that is given to `with_dumped` right?
> * In `Flamegraph.with_dumped`, this same value is passed to `with_`, and it's entirely ignored.
> 
> The documentation says that this value will be the name of the top-level frame in the graph but won't it always be ` "SYMBOLIC EXECUTION ROOT"`?

oh oops, ok so what's wrong is the documentation in Symex actually; it's not the name of the top-level frame, it's the name of the **file** that's created (notice `dest_file` in the code below). but it's a weird interface, so if you have other ideas id take it (maybe storing a `string option ref` in the flamegraph thing, and have an effect like `Flamegraph.set_name` that changes and locks the name?? and it would avoid having the argument in `'a effect_handler`?

```ocaml
  let with_dumped name f =
    match (Config.get ()).flamegraphs with
    | None -> with_ignored () f
    | Some dir ->
        let res, stack = with_ name f in
        let dest_file = Filename.concat dir name in
        dump dest_file stack;
        res
```

giltho:
Nah I'm happy with that for now, it makes sense I was just very confused by that doc haha
Let's just fix the doc and merge :)

PR #328, opened 2026-05-01
Consumption of non-consumable pure atoms via existential quantification

pcarrott:
This PR introduces two changes to the main Soteria library:
1. It fixes a previous bug that occurred during substitution of existential values in `bv_values`.
2. It extends `Logic.Asrt` with a new functor `Execute_partial` where `consume` is replaced with an alternative implementation: if only pure atoms remain to be consumed and neither is individually consumable, we create a new atom that existentially quantifies all unknown variables, making the atom consumable.

### DIFF
--- a/soteria/lib/bv_values/expr.ml
+++ b/soteria/lib/bv_values/expr.ml
@@ -65,27 +65,27 @@ module Subst = struct
         (v :: vs, s)
 
   and apply_bound ~missing_var s vs sv =
-    (* [max_var_ind] is the index of the freshest semantic variable used in
+    (* [max_var_ind] returns the index of the freshest semantic variable used in
        [sv], which we use to create new fresh variables for the existential. *)
+    let max_var_ind sv init =
+      Svalue.iter_vars sv
+      |> IterLabels.fold ~init ~f:(fun curr (var, _) ->
+          let var = Var.to_int var in
+          if var > curr then var else curr)
+    in
     let max_var_ind, s =
-      let incr_var_ind sv init =
-        Svalue.iter_vars sv
-        |> IterLabels.fold ~init ~f:(fun curr (var, _) ->
-            let var = Var.to_int var in
-            if var > curr then var else curr)
-      in
       Svalue.iter_vars sv
       |> Iter.filter (fun var -> not @@ List.mem var vs)
       |> IterLabels.fold ~init:(0, s) ~f:(fun (curr, s) (var, ty) ->
           let syn_var = mk_var var ty in
           match Raw_map.find_opt syn_var s with
-          | Some sv -> (incr_var_ind sv curr, s)
+          | Some sv -> (max_var_ind sv curr, s)
           | None ->
               (* We call [missing_var] here to ensure that our fresh variables
                  do not clash with other unknown variables and then update [s]
                  so that [missing_var] is only called once. *)
               let sv = missing_var var ty in
-              (incr_var_ind sv curr, Raw_map.add syn_var sv s))
+              (max_var_ind sv curr, Raw_map.add syn_var sv s))
     in
     (* [subst_with_fresh] is the new substitution which extends [s] by binding
        each variable in [vs] to a fresh semantic variable. [fresh_vs] is a list
--- a/soteria/lib/bv_values/expr.ml
+++ b/soteria/lib/bv_values/expr.ml
@@ -65,12 +65,12 @@ module Subst = struct
         (v :: vs, s)
 
   and apply_bound ~missing_var s vs sv =
-    (* [max_var_ind] is the index of the last semantic variable used in [sv],
-       which we use to create new fresh variables for the existential. *)
+    (* [max_var_ind] is the index of the freshest semantic variable used in
+       [sv], which we use to create new fresh variables for the existential. *)
     let max_var_ind, s =
-      let incr_var_ind sv curr =
+      let incr_var_ind sv init =
         Svalue.iter_vars sv
-        |> IterLabels.fold ~init:curr ~f:(fun curr (var, _) ->
+        |> IterLabels.fold ~init ~f:(fun curr (var, _) ->
             let var = Var.to_int var in
             if var > curr then var else curr)
       in
--- a/soteria/lib/bv_values/expr.ml
+++ b/soteria/lib/bv_values/expr.ml
@@ -65,16 +65,27 @@ module Subst = struct
         (v :: vs, s)
 
   and apply_bound ~missing_var s vs sv =
-    (* [max_var_ind] is the index of the last semantic variable, which we use to
-       create new fresh variables for the existential. *)
-    let max_var_ind =
-      Raw_map.fold
-        (fun _ binding curr ->
-          iter_vars binding
-          |> IterLabels.fold ~init:curr ~f:(fun curr (var, _) ->
-              let var = Var.to_int var in
-              if var > curr then var else curr))
-        s 0
+    (* [max_var_ind] is the index of the last semantic variable used in [sv],
+       which we use to create new fresh variables for the existential. *)
+    let max_var_ind, s =
+      let incr_var_ind sv curr =
+        Svalue.iter_vars sv
+        |> IterLabels.fold ~init:curr ~f:(fun curr (var, _) ->
+            let var = Var.to_int var in
+            if var > curr then var else curr)
+      in
+      Svalue.iter_vars sv
+      |> Iter.filter (fun var -> not @@ List.mem var vs)
+      |> IterLabels.fold ~init:(0, s) ~f:(fun (curr, s) (var, ty) ->
+          let syn_var = mk_var var ty in
+          match Raw_map.find_opt syn_var s with
+          | Some sv -> (incr_var_ind sv curr, s)
+          | None ->
+              (* We call [missing_var] here to ensure that our fresh variables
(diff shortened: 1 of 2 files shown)


soteria/lib/bv_values/expr.ml
```diff
@@ -53,39 +53,45 @@ module Subst = struct
...
+            (* [max_var_ind] is the index of the last semantic variable, which
+               we use to create new fresh variables for the existential. *)
+            let max_var_ind =
+              Raw_map.fold
+                (fun _ binding curr ->
+                  iter_vars binding
+                  |> IterLabels.fold ~init:curr ~f:(fun curr (var, _) ->
+                      let var = Var.to_int var in
+                      if var > curr then var else curr))
+                s 0
```
giltho: In doubt, you should also iterate over `sv` and accumulate all variables there I think.
I.e., if `sv` contains `x` but `x` is not bound in the substitution, `missing_var` could decide *not* to replace it. In which case, it needs to be counted as a possible semantic variable.

In general, you can massively over-approximate what counts as a semantic variable, the goal is to create fresh things there is no need for precision

soteria/lib/bv_values/expr.ml
```diff
@@ -53,39 +53,45 @@ module Subst = struct
             let vs, s = apply_list ~missing_var s vs in
             (Eval.eval_nop nop vs, s)
         | Exists (vs, sv) ->
```
giltho, line 55: Could you please extract a function called `apply_bound` which receives `s`, `vs` and `sv` and returns the final `s`, `vs` and `sv`?
First it would declutter this function, this specific pattern match is too big, and second we could trivially reuse it to add forall quantifiers.

Basically, this pattern should look like:
```ocaml
| Exists (vs, sv) ->(
   let ((vs, sv), s) = apply_bound s vs sv in
   (Svalue.Bool.mk_exists vs sv, s))
```

soteria/lib/logic/asrt.ml
```diff
@@ -83,31 +83,102 @@ module M (Symex : Symex.Base) = struct
...
+          Symex.Value.Expr.t
+
+        val conj : Symex.Value.Expr.t list -> Symex.Value.Expr.t
+      end) =
+  struct
+    include Execute (B)
+
+    (** Given a list of expressions for which [is_consumable] returns false,
+        returns a single expression for which [is_consumable] returns true. *)
+    let make_consumable (subst : Value.Expr.Subst.t) (exprs : Value.Expr.t list)
```
giltho: I don't like the name `make_consumable`.
`consumable` kind of means you're able to learn out of it. Here, you're giving up on learning.
How about `bind_free_vars_to_exist`?

soteria/lib/logic/asrt.ml
```diff
@@ -83,31 +83,102 @@ module M (Symex : Symex.Base) = struct
...
+           2>Substitution:@ %a@]@.@[<v 2>Atoms left:@ %a@]@]"
+          Value.Expr.Subst.pp subst (pp B.pp_syn) remaining];
+        Consumer.lfail @@ Value.of_bool false
+  end
+
+  (** [Execute_partial] contains the utilities to perform {e production} and
+      {e consumption} of a given assertion, using the consumer and producer of
+      the parameter [B]. In the presence of strictly-pure non-consumable atoms,
+      consumption will proceed by creating a new consumable atom via the
+      utilities of the parameter [Syn]. *)
```
giltho: This isn't informative enough I'd say. 
When writing documentation, try to ask yourself "who am I writing this for?"


```suggestion
  (** [Execute_partial] contains the utilities to perform {e production} and
      {e consumption} of a given assertion, using the consumer and producer of
      the parameter [B]. Unlike {!Execute}, the [consume] function may complete
      by learning {e partial substitutions}, i.e. substitutions that do not cover all free
      variables of the assertion being consumed.
      
      To achieve so, if there are leftover pure assertion to consume for which
      the free variables cannot be easily learned, we replace them by 
      existential (bound) variables and send that query to the solver. *)
```

giltho (approved):
Overall good work! I like how you implemented consume
I added some style comments, not much, but please address them before we merge :)

soteria/lib/bv_values/expr.ml
pcarrott: Technically, I think it suffices to look at the variables in `sv`. All other variables in `s` will not clash because they do not appear inside the existential. I've changed this part to only iterate over `sv`, not the entirety of `s`.

About `missing_var`, I think it's more complicated than that because `missing_var` can technically return everything? What I did in the most recent commit is call `missing_var` here (before actually performing the substitution) to know what variables are being generated. To avoid calling `missing_var` a second time during substitution, I immediately update `s` when I call it the first time.

Does this approach work? Or am I missing/complicating something?

soteria/lib/bv_values/expr.ml
pcarrott: Good idea! Fixed it in the most recent commits :)

soteria/lib/logic/asrt.ml
pcarrott: I was trying to keep the API a bit more abstract and simply expose a function that constructs a "consumable" expression from the leftovers. Basically, we are constructing an atom for which `consume_atom` should succeed. I don't know if there could be other approaches besides this existential hack, but I'm ok with being explicit about that if you prefer. The name `bind_free_vars_to_exists` is a bit lengthy but I can't think of anything better, so let's stick to that for now.

soteria/lib/logic/asrt.ml
pcarrott: Same comment as above. I was trying to keep things more abstract on purpose, but it might be best to be explicit about these details.

soteria/lib/bv_values/expr.ml
```diff
@@ -95,6 +64,57 @@ module Subst = struct
...
+    (* [max_var_ind] is the index of the last semantic variable used in [sv],
+       which we use to create new fresh variables for the existential. *)
+    let max_var_ind, s =
+      let incr_var_ind sv curr =
+        Svalue.iter_vars sv
+        |> IterLabels.fold ~init:curr ~f:(fun curr (var, _) ->
+            let var = Var.to_int var in
+            if var > curr then var else curr)
+      in
+      Svalue.iter_vars sv
```
N1ark, line 77: try avoiding nested functions when they don't need to be nested, they make the code pretty hard to read! also rename `incr_var_ind` to `max_var_ind` because we're not incrementing anything, it's a pure function
```suggestion
    let max_var_ind sv curr =
      Svalue.iter_vars sv
      |> IterLabels.fold ~init:curr ~f:(fun curr (var, _) ->
          let var = Var.to_int var in
          if var > curr then var else curr)
    in
    let max_var_ind, s =
      Svalue.iter_vars sv
```

soteria/lib/bv_values/expr.ml
```diff
@@ -95,6 +64,57 @@ module Subst = struct
...
+      |> Iter.filter (fun var -> not @@ List.mem var vs)
```
N1ark, line 78: i think this should be `fun (var, _) -> not @@ List.mem var bound_vars`, where `bound_vars = List.map fst vs`? i dont know if the type here matters, or if it can happen though

soteria/lib/logic/asrt.ml
```diff
@@ -83,31 +83,103 @@ module M (Symex : Symex.Base) = struct
...
+          (Var.t * 'a Value.ty) list -> Value.Expr.t -> Value.Expr.t
+
+        val conj : Value.Expr.t list -> Value.Expr.t
+      end) =
+  struct
+    include Execute (B)
+
+    (** Given a list of expressions for which [is_consumable] returns false,
+        returns the conjunction of all expressions where all free variables not
+        covered by [subst] are bound by an existential quantifier. *)
```
N1ark: `is_consumable` being false is irrelevant to this function so it's better to leave the doc more concise as the code is already quite hard to parse :') 
```suggestion
    (** Given a list of expressions,
        returns the conjunction of all expressions where all free variables not
        covered by [subst] are bound by an existential quantifier. *)
```

N1ark (approved):
small nits, thanks!

soteria/lib/bv_values/expr.ml
pcarrott: When I create the existential in `consume`, I call `List.mem` on the pair, not on just the `var`, so both functions are consistent with each other. But in general `mk_exists` does not really enforce this anywhere? I think it's technically possible to construct something like `∃ (x : ty1). (x : ty1) = (x : ty2)` with `mk_exists`, how should substitution handle these cases? Tbf this can even happen without the existential. I'll gladly change to what makes more sense tho. Should `x : ty1` and `x : ty2` be treated as the same variable or different variables?

soteria/lib/bv_values/expr.ml
pcarrott: Got it, thanks!

soteria/lib/bv_values/expr.ml
pcarrott: Actually, keys in `s` include the type of the variable, so `x : ty1` and `x : ty2` are treated as different variables and therefore have different bindings.

N1ark: perfect then if that's the case it's all ~good (i think it _could_ cause errors in smtlib but in practice let's assume this won't occur)

PR #329, opened 2026-05-01
Fix and test the import of Soteria as a library

giltho:
The cmdliner import was missing from Soteria itself.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -55,6 +55,7 @@
    (>= 0.5.0))
   (grace
    (>= 0.3.0))
+  zarith
   (odoc :with-doc)
   (mdx :with-test)
   (alcotest :with-test)
@@ -96,8 +97,16 @@
   soteria
   cerberus-lib
   zarith
+  cmdliner
+  iter
+  tsort
+  (progress
+   (>= 0.5.0))
   ppx_deriving
+  (ppx_subliner
+   (>= 0.2.1))
   fmt
+  dune-site
   (linol
    (= 0.10))
   (linol-eio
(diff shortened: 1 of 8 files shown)


giltho:
He that Claude inserting in commits

dune-project
```diff
@@ -94,15 +96,12 @@
...
   zarith
-  cmdliner
   ppx_deriving
   fmt
   (linol
    (= 0.10))
   (linol-eio
    (= 0.10))
-  (ppx_subliner
-   (>= 0.2.1))
```
N1ark, line 105: ?? we use that

giltho: It's already a dependency of Soteria

N1ark: it seems weird to not list our direct dependencies because they're also transitive dependencies; is that common in OCaml? like fair enough but i dont like that idea that if someday we top using it in soteria then soteria-rust/c imports break

giltho: I actually don't know. Since everything is expected to be in a single massive switch and everything is accessible to everyone... I'm used to specifying kind of minimal deps but maybe I'm wrong

N1ark: well yes to specifying minimal deps but like something you use should be a dep regardless of transitives imo.... if you can just undo this change im happy to merge

giltho:
Well, I'm either going to do it entirely or not do it at all. I made all direct dependencies explicit, and also realised a few deps were not directly used in soteria-c/rust but were declared there

giltho:
Auto-merge enable, see if you're happy with that last commit and just approve :)

N1ark (approved):
this is perfect thank you :)

PR #336, opened 2026-05-11
Recursive `ite` reductions, fix `if_sure`

N1ark:
- Add reductions to `ite` to make it subsitute the guard in the subterms. This avoids expressions of the form `a ? b : (a ? c : d)` and also simplifies e.g. `a ? b : a && c` => `a ? b : false` => `a && b`

  This required moving `Eval` to `Svalue`, so `ite` can take advantage of it; i think this is a relatively nice thing though. I also added a `?(subst=Fun.id)` param to `eval`, which allows substituting individual terms (not only variables)

- Fix two bugs in `if_sure`:
  1. the `then_` branch was executed with the `not guard` still in the PC, meaning that branch was always unsat
  2. the `else_` branch was executed with `guard` in the PC, which could prune branches; the else branch shouldn't have it's PC modified from the base case, because we have no guarantee as to whether the guard is true or not.
- Fix some `add_constraints ?simplified` calls in symex that had the wrong value in `?simplified`


N1ark:
!run-bv-fuzzer

N1ark:
!run-bv-fuzzer

giltho:
Love that the fuzzer caught your mistake

soteria-rust/test/cram/poly.t/run.t
```diff
@@ -8,21 +8,16 @@ Simple polymorphic function with no polymorphic values
   Compiling... done in <time>
   => Running trivial::trivial...
   note: trivial::trivial: done in <time>, ran 2 branches
-  PC 1: (((((0x0000000000000010 +ck V|1|) -ck extend[u61](extract[0-2](V|1|))) <u 0x0000000000000008) && (0b000 != extract[0-2](V|1|))) || ((0b000 == extract[0-2](V|1|)) || (0x0000000000000000 != ((0x0000000000000010 +ck V|1|) -ck extend[u61](extract[0-2](V|1|)))))) /\
```
giltho, line 11: Goddamn!!!

soteria-rust/lib/layout.ml
```diff
@@ -178,18 +179,28 @@ let rec layout_of (ty : Types.ty) : (t, 'e, 'f) Rustsymex.Result.t =
   | TFnDef _ -> ok (mk_concrete ~size:0 ~align:1 ~fields:Primitive ())
   (* Type variables : non-deterministically generate a layout *)
   | TVar (Free _) ->
+      let open Usize in
       (* FIXME: we need to scope these type variables, as the T in foo<T> and in
          bar<T> are "different" Ts. *)
       let* size = nondet (Typed.t_usize ()) in
-      let* () = assume Usize.[ 0s <=$@ size; size <$@ 1024s ] in
+      (* We assume the size is at most 1024 bytes, which is a reasonable upper
+         bound for now. *)
```
giltho, line 187: Why is this reasonable again?

giltho: And more importantly why is it necessary?

soteria/lib/bv_values/svalue.ml
```diff
@@ -2088,6 +2169,10 @@ and BitVec : BitVec = struct
       when if signed then msb_of v1 + msb_of v2 < size_of v1.node.ty - 2
            else msb_of v1 + msb_of v2 < size_of v1.node.ty - 1 ->
         Bool.v_false
+    | Ite (g1, t1, e1), Ite (g2, t2, e2) when equal g1 g2 ->
+        let overflow_then = mul_overflows ~signed t1 t2 in
+        let overflow_else = mul_overflows ~signed e1 e2 in
+        Bool.ite g1 overflow_then overflow_else
```
giltho, line 2175: Can't this be lifted for any binop? E.g. make it part of `mk_binop` or something

soteria/lib/bv_values/svalue.ml
```diff
@@ -2218,9 +2307,119 @@ and Float : Float = struct
...
+      let res = eval ~force ~eval_var ~subst x in
+      if equal res x then res else subst res
+    in
+    match x.node.kind with
+    | Var v -> eval_var x v x.node.ty
+    | Bool _ | Float _ | BitVec _ -> x
+    | Ptr (l, o) ->
+        let nl = eval l in
+        let no = eval o in
+        if (not force) && equal l nl && equal o no then x else Ptr.mk nl no
```
giltho, line 2374: I don't quite understand how subst is used, why is it only used on the children? What if the subst result needs to be simplified in turn?

giltho (commented):
I need to think a bit deeper about whether this can be simplified, it's a bit heavy...

soteria-rust/lib/layout.ml
N1ark: it's necessary bc if the size is unbounded (i.e. `isize::MAX` at most), then the second you try using that type in a struct, or for an array or anything, you get a layout error due to an overflow. `1024` is just a dummy size that basically means you're unlikely to ever get an overflow when doing layout calculations, but it's big enough that you could catch something interesting (maybe)

soteria/lib/bv_values/svalue.ml
N1ark: hmmm yes! i can do that if youd prefer it; i dont mind the noise here but either is fine, its honestly kind of overkill

soteria/lib/bv_values/svalue.ml
N1ark: wdym "only used on the children"? it's used everywhere (note there is also a `subst` in the initial call to `eval`)

i didnt re-eval the subst result just to avoid nasty loops, its just a small trick/utility but it shouldn't worsen performance too much... all of this _is_ a little ad-hoc, im happy to change it to your taste

giltho: I mean that, say my subst is: `s = { (a \/ b) : true }`
But then `eval ~subst:s  (a \/ b)` will do nothing, I think?

While, in contract, `eval ~subst:s ((a \/ b) \/ c)` will evaluate to `true`

N1ark: again, it _is_ used everwhere :') check below

```ocaml
  let eval ?(force = false) ?(eval_var : t -> Var.t -> ty -> t = fun x _ _ -> x)
      ?(subst : t -> t = Fun.id) (x : t) : t =
    try eval ~force ~eval_var ~subst (subst x) with Division_by_zero -> x
```

giltho: Ah I did look and didn't see `(subst x)`

Hmm, not sure what I think about all this yet. I feel like there's a slightly cleaner implementation to have

PR #337, opened 2026-05-11
Deploy linux package

giltho:
Deploy linux packages as well as macos packages. Let's not merge this until I've tested I've updated cargo-soteria (for the artifact name changes), and tested it inside a docker vm.

This opened a whole can of worm about reproducibility of our tests over every platform.
I have made the following modifications:
- `--hide-unstable` also uses a universal profile now (no-color & no-utf8)
- added `__builtin_bswap16` to soteria-c.h, that offset all indexes of symbols, because Cerberus gives a unique indentifier, and this was parsed very early, so it changed all values... 
- to stop all C tests from changing every time I modify soteria-c.h, I print the identifier through `pp_unstable`

### DIFF
--- a/.github/workflows/test-packages.yml
+++ b/.github/workflows/test-packages.yml
@@ -43,6 +43,7 @@ jobs:
       - name: Setup environment variables
         run: |
           PACKAGE_DIR="${GITHUB_WORKSPACE}/soteria-c-package"
+          echo "PACKAGE_DIR=${PACKAGE_DIR}" >> $GITHUB_ENV
           echo "CERB_INSTALL_PREFIX=${PACKAGE_DIR}" >> $GITHUB_ENV
           echo "SOTERIA_Z3_PATH=${PACKAGE_DIR}/bin/z3" >> $GITHUB_ENV
           echo "SOTERIA_AUTO_INCLUDE_PATH=${PACKAGE_DIR}/lib/soteria-c" >> $GITHUB_ENV
@@ -120,6 +121,7 @@ jobs:
       - name: Setup environment variables
         run: |
           PACKAGE_DIR="${GITHUB_WORKSPACE}/soteria-rust-package"
+          echo "PACKAGE_DIR=${PACKAGE_DIR}" >> $GITHUB_ENV
           echo "SOTERIA_Z3_PATH=${PACKAGE_DIR}/bin/z3" >> $GITHUB_ENV
           echo "SOTERIA_OBOL_PATH=${PACKAGE_DIR}/bin/obol" >> $GITHUB_ENV
           echo "SOTERIA_CHARON_PATH=${PACKAGE_DIR}/bin/charon" >> $GITHUB_ENV
--- a/.github/workflows/test-packages.yml
+++ b/.github/workflows/test-packages.yml
@@ -14,9 +14,17 @@ jobs:
     name: Test soteria-c Package
     strategy:
       matrix:
-        operating-system: [macos-latest]
+        operating-system: [macos-latest, ubuntu-latest]
     runs-on: ${{ matrix.operating-system }}
     steps:
+      - name: Set platform artifact name
+        id: platform
+        run: |
+          if [ "${{ runner.os }}" = "macOS" ]; then
+            echo "name=macos-arm64" >> $GITHUB_OUTPUT
+          else
+            echo "name=linux-x86_64" >> $GITHUB_OUTPUT
+          fi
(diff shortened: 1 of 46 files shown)


packaging/common_lib.ml
```diff
@@ -11,13 +11,24 @@ let dest_dir_arg position =
...
+        |. read_all
+        |> eval
+        |> print_string
+    | _ ->
+        (* Linux: ldd output is " libfoo.so => /path/to/libfoo.so (0xaddr)" *)
+        run "ldd" [ exe ]
+        |. run "awk" [ "/=>/ {print $3}" ]
+        |. read_all
+        |> eval
+        |> print_string
```
N1ark: can you put a string there to match and have a `_ -> failwith "unsupported OS"` instead as a catch all? this seems weird

.github/workflows/test-packages.yml
```diff
@@ -41,6 +50,17 @@ jobs:
...
+        if: runner.os == 'Linux'
+        run: |
+          PACKAGE_DIR="${GITHUB_WORKSPACE}/soteria-c-package"
+          echo "CERB_INSTALL_PREFIX=${PACKAGE_DIR}" >> $GITHUB_ENV
+          echo "LD_LIBRARY_PATH=${PACKAGE_DIR}/lib:${LD_LIBRARY_PATH}" >> $GITHUB_ENV
+          echo "SOTERIA_Z3_PATH=${PACKAGE_DIR}/bin/z3" >> $GITHUB_ENV
+          echo "SOTERIA_AUTO_INCLUDE_PATH=${PACKAGE_DIR}/lib/soteria-c" >> $GITHUB_ENV
+          echo "SOTERIA_C=${PACKAGE_DIR}/bin/soteria-c" >> $GITHUB_ENV
+          echo "SOTERIA_SOLVER_TIMEOUT=100" >> $GITHUB_ENV
+          echo "SOTERIA_USE_CERB_HEADERS=true" >> $GITHUB_ENV
```
N1ark: this + the bit below feel like a lot of duplication... what if you merge both blocks and just have an if-elseif-else over the os to set the ld lib path?

N1ark (commented):
cool we got support for one more OS:)

giltho:
What do we do with this @N1ark ?
https://github.com/soteria-tools/soteria/actions/runs/25695807902/job/75443608316

image

N1ark:
oh funny! try modifying these tests and passing a --target flag (and take it as an opportunity to duplicate the tests to test it both w linux and macos), that should work maybe

giltho:
> oh funny! try modifying these tests and passing a --target flag (and take it as an opportunity to duplicate the tests to test it both w linux and macos), that should work maybe

Very annoying, it means one needs both targets installed to get that test working...
I'm going to create `linux.t` and `macOS.t`

N1ark:
> Very annoying, it means one needs both targets installed to get that test working... I'm going to create `linux.t` and `macOS.t`

i mean yeah but they're tests, I think it's ok to expect more of the testing environment; typically Charon expects all of these: x86_64-apple-darwin, x86_64-pc-windows-msvc, aarch64-apple-darwin, i686-unknown-linux-gnu, powerpc64-unknown-linux-gnu, riscv64gc-unknown-none-elf

giltho:
Isn't our CI slow enough? ☠️
But sure I can do that.

N1ark:
> Isn't our CI slow enough? ☠️ But sure I can do that.

The rust toolchain is cached, so it shouldn't change anything? You don't need to run the CI on multiple targets; Rust supports cross-target compilation

soteria/lib/logs/profile.ml
```diff
@@ -28,16 +28,21 @@ let supports_utf8_from_env () =
...
 
-let mk_profile () =
-  let utf8 = supports_utf8_from_env () in
-  let color = color_from_env () in
-  { color; utf8 }
+(* Profile that should be supported on all platforms. *)
+let universal = { color = false; utf8 = false }
+
+let mk_profile ~hide_unstable () =
+  if hide_unstable then universal
```
N1ark, line 35: can you update this in the logs tutorial

soteria/lib/logs/config.ml
N1ark: here too; mention it disables utf8 and color output

soteria/lib/logs/profile.ml
giltho: Good catch!

N1ark (approved):

PR #339, opened 2026-05-12
Remove the VSCode plugin for Soteria-C

giltho:
As discussed:
- We don't use it
- Given the recent npm security issues, it's safer

Let's note that the code remains on this branch and can always be reverted if we want it back


N1ark (approved):
farewell

PR #340, opened 2026-05-15
Bump Charon, fixes, Rust improvements

### LINKED ISSUE #285: Check decayed address for alignment checks
Currently we don't support this sort of low-level code ([source](https://github.com/rust-lang/miri/blob/4080db1cfdbc244282060cf0e9a02d6da7e0ec2a/tests/pass/align.rs#L3-L12))

```rust
let x = &mut [0u8; 3];
let base_addr = x as *mut _ as usize;
let base_addr_aligned = if base_addr % 2 == 0 { base_addr } else { base_addr + 1 };
let u16_ptr = base_addr_aligned as *mut u16;
unsafe {
    *u16_ptr = 2;
}
```

Our alignment check considers the pointer's allocation's alignment and it's offset separately:
https://github.com/soteria-tools/soteria/blob/97fee7d7a4ae1f19896453d6f5b35ee2c71d9cde/soteria-rust/lib/sptr.ml#L323-L329

Although we should not generate an address for all allocations, we can, if the first alignment check fails **and the allocation's address has been decayed**, get the address and check alignment with that address + offset. This should enable these sorts of operations without decaying more pointers than necessary.

N1ark:
- Bump Charon, and switch to using the Postcard serialization format by default (Obol now only supports that one).
- Check addresses when checking alignment; closes #285
- Support the `exit` libc call (which enables `std::process::exit`)
- Fix `if%sure` being completely wrong

And for my own pleasure: a comparison chart of running Soteria Rust against Kani and Miri on their test suites, before and after the switch to postcard format :) 

| Before | After |
| ----- | ----- |
|  |  |

> Superimposed (lighter color is old)

### DIFF
--- a/soteria-rust/lib/state/tree_state.ml
+++ b/soteria-rust/lib/state/tree_state.ml
@@ -490,7 +490,19 @@ module Make (Borrows : Tree_borrows.T) = struct
     let is_aligned =
       ofs %@ exp_align ==@ Usize.(0s) &&@ (align %@ exp_align ==@ Usize.(0s))
     in
-    assert_or_error is_aligned (`MisalignedPointer (exp_align, align, ofs))
+    if%sat Typed.not is_aligned then
+      (* If we can't guarantee the pointer is aligned from the type's alignment,
+         we try decaying it and seeing if its (symbolic) address **implies** its
+         alignment. Note this is not OX sound; we only take the ok path if the
+         pointer **must** be aligned, rather than also when it may be aligned.
+
+         This avoids going through extra branches, as a pointer can pretty much
+         always be aligned; what matters most is whether it is guaranteed to be
+         aligned. *)
+      let* address = with_pointers_sym @@ Sptr_base.decay ptr in
+      if%sure address %@ exp_align ==@ Usize.(0s) then Result.ok ()
+      else Result.error (`MisalignedPointer (exp_align, align, ofs))
+    else Result.ok ()
 
   and check_non_dangling_untyped ((ptr : Sptr_base.t), _) size =
     if%sat size ==@ Usize.(0s) then Result.ok ()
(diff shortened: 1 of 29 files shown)


soteria-rust/lib/state/tree_state.ml
```diff
@@ -490,7 +490,19 @@ module Make (Borrows : Tree_borrows.T) = struct
...
+         we try decaying it and seeing if its (symbolic) address **implies** its
+         alignment. Note this is not OX sound; we only take the ok path if the
+         pointer **must** be aligned, rather than also when it may be aligned.
+
+         This avoids going through extra branches, as a pointer can pretty much
+         always be aligned; what matters most is whether it is guaranteed to be
+         aligned. *)
+      let* address = with_pointers_sym @@ Sptr_base.decay ptr in
+      if%sure address %@ exp_align ==@ Usize.(0s) then Result.ok ()
+      else Result.error (`MisalignedPointer (exp_align, align, ofs))
```
giltho, line 504: Wouldn't `assert_or_error` + fail-fast be equivalent though?

Maybe we should have `assert_or_error ~fail_fast:true`?

giltho (approved):
Absolutely banger for postcard!
We need to investigate how much work a `charon-server` and `obol-server` would be

soteria-rust/lib/state/tree_state.ml
N1ark: it's not fail fast because if the misaligned pointer error is raised you still explore the rest of the branches; what you want is more like `assert_sure_or_error` or `assert_or_error ~ux` i guess

giltho: It's "fail fast for this specific branch" but yeah

PR #345, opened 2026-05-17
Make `Symex.run` return a syntactic PC 

N1ark:
Updated the tutorial to mention it


soteria/lib/symex/symex.ml
```diff
@@ -1030,12 +1027,14 @@ module Make (Sol : Solver.Mutable_incremental) :
     let@ () = Give_up.with_give_up_raising in
     f ()
 
-  let run_iter ~mode (iter : 'a t) : ('a * Value.(sbool t) list) Iter.t =
+  let solver_as_exprs () = Solver.as_values () |> List.map Value.Expr.of_value
```
giltho: Solver shouldn't expose `as_values` I think, it should only expose `as_exprs`.

giltho (approved):
LGTM, I'd change the solver interface, that's it

soteria/lib/symex/symex.ml
N1ark: thought about it but wasn't sure, thanks :)

PR #346, opened 2026-05-17
Auto-sync website API doc

### LINKED ISSUE #333: Move website to website.
Would be good to deploy the documentation to the soteria-tools website, using markdown and [this approach](https://sancho.dev/blog/ocaml-documentation-as-markdown).

We could customise navigation and style in a lot nicer way

N1ark:
Add a new CI step; when pushing to main, the JSON docs are pushed to an orphan branch `docs-json`, and a workflow is triggered on the website to pull that branch with the updated docs

Closes #333 
Website docs: https://soteria-tools.com/docs/library

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -167,5 +167,17 @@ jobs:
         # We do this on linux because linux runners are cheaper.
         if: runner.os == 'linux'
 
+      - name: Build JSON documentation
+        run: make doc-json
+        if: runner.os == 'linux'
+
+      - name: Upload JSON documentation artifact
+        uses: actions/upload-artifact@v7
+        with:
+          name: odoc-json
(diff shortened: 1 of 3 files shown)


giltho:
Huh
Why... Don't you pull the workflow artifact from the website workflow instead of pushing to an orphan branch? What will happen if it runs twice in parallel? I think you can just pass arguments to workflows, you could give the artifact id or something?

N1ark:
You can't pull artifacts between repos :/
What I originally had was that this repo would commit to website/main with the new API docs, but then it means if the CI script gets tampered with someone could use a write-access token on the website. So instead I do it pull-based, which means the worst that can happen on this side is that the workflow isnt triggered

giltho (approved):

PR #354, opened 2026-05-25
Improve aliasing logic in stubs

giltho:
## Summary

  - Rework how `aliases` work in `stubs.json`: an alias is now a list (`aliases: [...]`) of extra `fn_pats` keys that route to the same stub variant as the canonical `pattern`. Signature resolution is driven only by the `pattern`.
  - Dedupe `fn_pats` entries so a path appearing as both a pattern and another entry's alias yields a single mapping.
  - Drop the now-redundant standalone `std::panicking::begin_panic` and `std::io::stdio::_print` entries; merge them into the aliases of their respective canonical patterns.
  - Fix the JSON schema: replace the old `alias` (string|array) definition with a list-only `aliases`, and remove an invalid `"required": true` inside the `patterns` property (it now validates against jsonschema spec draft 2020-12).
  - Quote the `tuple[UniqueType, UniqueType]` forward reference so `stubs.py` parses on Python 3.11 (it only worked on Python 3.14+)
  - Rename the corresponding stub functions in `optim/impl.ml` (`stdio__print` → `_print`, `panicking_begin_panic` → `begin_panic`) since `unique_name` no longer needs to disambiguate them.

### DIFF
--- a/soteria-rust/lib/builtins/optim/impl.ml
+++ b/soteria-rust/lib/builtins/optim/impl.ml
@@ -114,13 +114,13 @@ module M (StateM : State.StateM.S) : Intf.M(StateM).S = struct
   let panic_fmt ~fmt:_ = do_panic ()
   let panic_nounwind_fmt ~fmt:_ ~force_no_backtrace:_ = do_panic ()
 
-  let panicking_begin_panic ~m:_ ~msg =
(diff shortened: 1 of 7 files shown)


N1ark (approved):
nice thank you :)

PR #356, opened 2026-05-25
BtreeSet sort benchmark + trivial inequalities

giltho:
We didn't have any benchmark that entailed a lot of branches so I added one. I asked Claude to make it faster and it found a very nice improvement. It could eventually be replaced by a relational domain, but for now it's cute.

| metric | before | after | Δ |
| --- | --- | --- | --- |
| wall-clock (hyperfine, 8 runs) | 7.268 s ± 0.176 | 6.820 s ± 0.036 | −6.2% |
| Z3 queries | 352 | 119 | −66 % |
| Z3 cumulative time | 685 ms | 207 ms | −70 % |
| branches | 75 | 75 | unchanged |


giltho:
!run-benchmarks

soteria/lib/bv_values/bv_solver.ml
```diff
@@ -2,6 +2,36 @@ open Soteria_std
...
+    when qs = ps && equal qa pa && equal qb pb ->
+      Some true
+  (* [a < b] in PC implies ~[b <= a] and ~[b < a] *)
+  | Binop ((Lt qs | Leq qs), qa, qb), Binop (Lt ps, pa, pb)
+    when qs = ps && equal qa pb && equal qb pa ->
+      Some false
+  (* [a <= b] in PC implies ~[b < a] *)
+  | Binop (Lt qs, qa, qb), Binop (Leq ps, pa, pb)
+    when qs = ps && equal qa pb && equal qb pa ->
+      Some false
```
N1ark: while we're at it
```suggestion
  | Binop (Leq qs, qa, qb), Binop (Lt ps, pa, pb)
    when Bool.equal qs ps && equal qa pa && equal qb pb ->
      Some true
  (* [a < b] in PC implies ~[b <= a] and ~[b < a] *)
  | Binop ((Lt qs | Leq qs), qa, qb), Binop (Lt ps, pa, pb)
    when Bool.equal qs ps && equal qa pb && equal qb pa ->
      Some false
  (* [a <= b] in PC implies ~[b < a] *)
  | Binop (Lt qs, qa, qb), Binop (Leq ps, pa, pb)
    when Bool.equal qs ps && equal qa pb && equal qb pa ->
      Some false
```

soteria/lib/bv_values/bv_solver.ml
```diff
@@ -2,6 +2,36 @@ open Soteria_std
 open Logs.Import
 module Var = Svalue.Var
 
+(** Returns [Some true] if PC slot [pc] implies query [q], [Some false] if [pc]
+    implies the negation of [q], and [None] otherwise. Used to suppress
+    redundant ordering constraints (e.g. [a <= b] becomes trivially true once
+    [a < b] is in the PC). *)
+let implies_or_contradicts (q : Svalue.t) (pc : Svalue.t) : bool option =
```
N1ark:
```suggestion
let[@inline] implies_or_contradicts (q : Svalue.t) (pc : Svalue.t) : bool option =
```

soteria/lib/bv_values/bv_solver.ml
```diff
@@ -77,7 +107,7 @@ struct
       find_map t (fun value ->
           if Typed.equal value v then Some true
           else if Typed.equal value neg_v then Some false
-          else None)
+          else implies_or_contradicts (Typed.untyped v) (Typed.untyped value))
```
N1ark: maybe if we do that we might as well also add these two checks (for equality/negated equality) in there...? idk

N1ark (approved):
heh it's cute, why not 


opinions on killing the incremental solver? it's kind of a maintenance burden, we might as well just keep it in tiny_values

soteria/lib/bv_values/bv_solver.ml
giltho: When something types as bool (or any primitive type), `=` just monomorphises at compilation. This is actually more readable, so I prefer not to change it

soteria/lib/bv_values/bv_solver.ml
giltho: Yes, and we can just compute the negation a single time as well probably

PR #359, opened 2026-05-28
Dependency upgrades & cleanup rust files

giltho:
- Upgrade to 5.4.1
- Upgrade to dune 3.23.1 (don't use 3.23.0 it has an annoying bug)
- Remove need to check for opam files in CI (dune test does it now). Also, they should be rebuilt using `dune build @opam --auto-promote` if needed
- `make cleanup-rust-tests` to cleanup annoying files
- `dune test` can now run even with tests haven't been cleaned up


N1ark (approved):
thanks :)

PR #363, opened 2026-06-01
Bump Charon

N1ark:
(draft bc im benchmarking)

Bump Charon to the latest (2026-05-31!!!) toolchain version (talk about cutting edge...)

The main change is adding pattern type support (used e.g. in `NonNull` and `NonZero`); this allows us to also get rid of the horrendous hack i had to parse the rustc `rustc_layout_scalar_valid_range_*` attributes <3 

we also now properly check the abi-string of function pointers to determine their own abi compatibility, and we support repr(transparent) too :)

PR #366, opened 2026-06-03
Cleaner transmutes

N1ark:
Implement a proper transmutation in `State`, that doesn't require allocating a block, reading from it and then freeing it. This is much better for performance, as it avoids a bunch of extra checks we don't care about (e.g. alignment of the "fake" block, or checking the used pointer is correct)

### DIFF
--- a/soteria-rust/lib/builtins/core.ml
+++ b/soteria-rust/lib/builtins/core.ml
@@ -181,25 +181,6 @@ module M (StateM : State.StateM.S) = struct
           "Unexpected operation or value in eval_ptr_binop: %a, %a, %a"
           Expressions.pp_binop op pp_rust_val l pp_rust_val r
 
-  let transmute ~from_ty ~to_ty v =
-    (* Some fun details:
-     *  - we need to use [get_state] and re-use the current state, rather than
-     *    using an empty state, because if the [load] does any reference
-     *    validity checks we need the current state to have these addresses!
-     *  - we need to take the max of either types for the alignment, to ensure
-     *    that transmuting e.g. from [u16; 2] to (u32) works. *)
-    [%l.debug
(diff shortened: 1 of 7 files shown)


N1ark:
!run-benchmarks

N1ark:
so taking into account [the previous regression](https://github.com/soteria-tools/soteria/pull/363#issuecomment-4610723463), this leaves us at:

| Benchmark | old main | charon update | this PR | Δ (old→this) |
  | --- | ---: | ---: | ---: | ---: |
  | rust-file: write-a-lot | 80.2 ms | 82.0 ms | 80.2 ms | 0.0% |
  | rust-file: write-a-lot (for loop) | 565.3 ms | 583.4 ms | 586.0 ms | 🔴 +3.7% |
  | rust-file: ctpop | 1.870 s | 1.897 s | 1.897 s | 🔴 +1.4% |
  | rust-file: array_init (rust) | 214.0 ms | 223.4 ms | 221.1 ms | 🔴 +3.3% |
  | rust-file: btreeset sort (size 4) | 6.313 s | 6.876 s | 6.652 s | 🔴 +5.4% |
  | c: array_init (c) | 1.961 s | 1.962 s | 1.972 s | 🔴 +0.6% |
  | c-capture-db: Collections-C | 9.866 s | 9.738 s | 9.703 s | 🟢 -1.7% |

we didn't fully recover performance but we're at +5 rather than +10, i don't think i can do much better, patterns just require more work but it's "good" in the sense that we now properly check that `NonNull` is not null :P

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
i guess that's as good as it gets </3 oh well i'm okay with this, i think there are big wins to have elsewhere if we really try

giltho (approved):

PR #368, opened 2026-06-04
Detect writes to read-only memory

N1ark:
We now properly track the source of allocations!
This means
- I could simplify leak checks significantly (and avoid false positives there)
- We now can detect and report writes to read-only memory (e.g. `const`s, or `&'static str`)

Originally I did all this because I wanted to remove both `update_ref_tys_in` and `ref_tys_in` (which play awfuly with what im doing in a future PR). Unfortunately I failed there but I managed to simplify it enough that it's not too bad anymore regardless.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,8 +8,8 @@ env:
   OCAML_VERSION: 5.4.1
   # [versionsync: OBOL_REPO=soteria-tools/obol]
   OBOL_REPO: soteria-tools/obol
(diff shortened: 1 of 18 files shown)


N1ark:
!run-benchmarks I suspect there will be a degradation because of the additional checks done for writes, my hope is that it's a small cost

N1ark:
i'll take it : )

giltho (approved):
cool

PR #370, opened 2026-06-05
Implement the `caller_location` intrinsic

N1ark:
Implement `caller_location` -- we used to have it as a stub that returned the null pointer, leading to null dereferences that were unintended.

Also implement `panic_impl`, which is an external function used by Rust's runtime; really it should resolve to the `#[panic_handler]` function but we don't translate attributes sufficiently in Charon/Obol for this (yet!!)


giltho (approved):
Niceeeeeeeeee

PR #373, opened 2026-06-08
Factor out logic of python scripts into soteria_utils.py

giltho:
We're starting to have quite a lot of script that reimplement printing logic etc. This PR factors all of it out, and then runs ty and ruff and fixes all lint issues.


giltho:
!run-benchmarks

(necessary because the PR changes the benchmarks scripts)

N1ark (approved):
nice

PR #374, opened 2026-06-09
Optimise struct and array access in the store

### LINKED ISSUE #216: Store optimisation with fields
Can we expand what we did for #162, and make it work for fields? Basically a "lazy address" can also carry a list of projections that we apply when reading/writing the value, by unwrapping the `Rust_val` appropriately and modifying whatever cell we must change.

Note this **does not hold for union fields**

An example of code we'd want to run with no heap access:

```rust
struct A { b : B }
struct B { x : u32 }

fn main() {
  let a = A { b : B { x : 12 } };
  let x = a.b.x;
}
```

While we're at it I wonder if we can't just replace `lazy_ptr` by `Charon.Expressions.place`, with the condition that there is no `Deref` or `Subslice` in the projection basically.

giltho:
As the title says. At least on the binary_tree benchmark, this is a 2x speedup

Closes #216


giltho (commented):
!run-benchmarks

giltho:
!run-benchmarks

soteria-rust/test/cram/simple.t/run.t
```diff
@@ -600,3 +600,19 @@ Ensure we implement the caller_location intrinsic correctly; this used to cause
...
+
+Test that taking a reference to a ZST doesn't allocate it on the heap; the reference is a dangling pointer, so the value stays in the store.
+  $ soteria-rust exec zst_ref.rs
+  Compiling... done in <time>
+  => Running zst_ref::main...
+  note: zst_ref::main: done in <time>, ran 1 branch
+  PC 1: empty
+  
+
+Test that indexing arrays with a constant index does not allocate; the value is updated in place in the store.
```
N1ark, line 653: dont you then want to add a `--stats stdout` or something to show that's the case? or output stats to a file and using `jq` to read it idc

giltho:
!run-benchmarks

N1ark:
wut why is perf significantly not as good

soteria-rust/lib/interp.ml
```diff
@@ -499,7 +499,10 @@ module Make (StateImpl : State.S) = struct
         match Store.try_load sp store with
         | Some (Value v) ->
             Soteria.Stats.As_ctx.incr StatKeys.loads_from_store;
-            ok v
+            (* a value read from the store skips the heap decoder, so we still
+               check its validity here, as [State.load] would *)
+            let+ () = State.check_validity ~check_refs:true ty v in
+            v
```
N1ark: this is really really really weird to me; if the store optim is used, then necessarily the value was written at some point and this without transmuting, but then that value itself should have had its validity checked. it seems we are instead not detecting somewhere that we're creating an invalid value; could it be in `zst_value`?

giltho:
!run-benchmarks

N1ark:
i havent read through the code, but you didn't stop allocating ZSTs; if a call to `State.alloc` goes through with a ZST, State will still allocate a block. what we need to fix is that any calls to `alloc_ty`/`alloc_untyped` with a size of zero are caught and return a dangling pointer

this can (and should) be done in a future PR though, this one is already quite technical and tricky

giltho:
It wasn't noise :(

giltho:
!run-benchmarks

soteria-rust/test/cram/simple.t/run.t
```diff
@@ -600,3 +600,31 @@ Ensure we implement the caller_location intrinsic correctly; this used to cause
...
+Test that our check_allocs test works appropriately by checking a case where we expect an allocation, and one where we don't.
+  $ check_allocs box.rs 0
+  Compiling... done in <time>
+  => Running box::main...
+  note: box::main: done in <time>, ran 1 branch
+  PC 1: (0x0000000000000004 <=u V|1|) /\ (V|1| <=u 0x7ffffffffffffffa) /\
+        (0b00 == extract[0-1](V|1|))
+  
+  check_allocs: expected '0', got '2'
+  [1]
```
N1ark: supposedly this should have one alloc, not zero

soteria-rust/lib/value_codec.ml
```diff
@@ -295,6 +295,13 @@ struct
         let discr = BV.of_literal variant.discriminant in
         Enum (discr, fields)
     | Enum _, _ -> failwith "decode: expected enum type for enum layout"
+
+  (** Returns the unique value of the type [zst]. Only safe to call if [zst] is
+      zero-sized. *)
+  let zst_value ty =
+    let error = fun _ -> failwith "Decoding a ZST that is not a ZST" in
+    ParserMonad.parse ~handler:error ~get_all:error
+      (decode ~meta:Thin ~offset:Usize.(0s) ty)
```
N1ark: thinking about this and i don't like it... i wonder if we can't just do this and call it a day???
```ocaml
let ptr = Sptr.at_location align in
let zst = State.load ty ptr in 
...
```

soteria-rust/lib/store.ml
```diff
@@ -6,25 +6,97 @@ module Map = Stdlib.Map.Make (struct
...
+
+  (** Local root of the place *)
+  let rec root p =
+    match p.kind with
+    | Local v -> v
+    | Field (base, _, _) | Index (base, _) -> root base
+
+  (** The root value [v] with the value at [sp] replaced by [f] applied to it;
+      [None] if not navigable. *)
+  let rec update_val { kind; _ } ~f v =
```
N1ark, line 38: what is `sp` here? the first argument?

```suggestion
  (** [update_val sp ~f v] is the value [v] with the value at [sp] replaced by applying [f] to it;
      [None] if not navigable. *)
  let rec update_val { kind; _ } ~f v =
```

soteria-rust/lib/rust_val.ml
```diff
@@ -246,3 +246,19 @@ let mk_struct ~ty fields =
...
+(** [v] with its [i]th field replaced by [f] applied to the old value; [None] if
+    [v] is not a navigable aggregate or [i] is out of range. *)
+let update_field i ~f v =
+  let update vs =
+    match List.nth_opt vs i with
+    | Some old ->
+        Option.map
+          (fun field -> List.mapi (fun j x -> if j = i then field else x) vs)
+          (f old)
+    | None -> None
```
N1ark: while you're at it with micro-optimisations, i'd define a function `List.update_at` and use that, which avoids re-allocating the rest of the tail
```ocaml
let[@tail_mod_cons] rec update_at f n = function
  | x :: rest -> if n = 0 then f x :: rest else x :: update_at f (n - 1) rest
  | [] -> failwith "Index out of bounds in update_at"
```

soteria-rust/lib/rust_val.ml
```diff
@@ -246,3 +246,19 @@ let mk_struct ~ty fields =
...
+  in
+  match v with
+  | Tuple vs -> Option.map (fun vs -> Tuple vs) (update vs)
+  | Enum (d, vs) -> Option.map (fun vs -> Enum (d, vs)) (update vs)
+  | _ -> None
```
N1ark: i would prefer we remove the options here; `update_field` must only be called on tuples/enums, so if we don't we're doing something wrong and we should most definitely catch it loudly

soteria-rust/lib/store.ml
```diff
@@ -6,25 +6,97 @@ module Map = Stdlib.Map.Make (struct
...
+  [@@deriving show { with_path = false }]
+
+  let as_value = function { kind = Value v; _ } -> Some v | _ -> None
+
+  let bind_value (f : 'a Rust_val.t -> 'a kind option) (x : 'a kind option) :
+      'a kind option =
+    match x with
+    | Some (Value v) -> f v
+    | Some Uninit -> Some Uninit
+    | Some Dead -> Some Dead
```
N1ark: while we're at it, i guess
```suggestion
    | (Some Uninit | Some Dead) as v -> v
```

soteria-rust/lib/store.ml
```diff
@@ -48,3 +120,49 @@ let get_ty sym t =
...
+      try_load base store
+      |> bind_value @@ function
+         | Tuple vs | Enum (_, vs) ->
+             Option.map (fun v -> Value v) (List.nth_opt vs field_idx)
+         | _ -> None)
+  | Index (base, idx) -> (
+      try_load base store
+      |> bind_value @@ function
+         | Tuple vs -> Option.map (fun v -> Value v) (List.nth_opt vs idx)
+         | _ -> None)
```
N1ark: same here; we don't want the `_ -> None` case because it's dangerous and we have the guarantee things are well-typed

soteria-rust/lib/state/tree_state.ml
```diff
@@ -9,6 +9,15 @@ open Common
...
 open Sptr
 
+module Stat_keys = struct
+  let allocs = "soteria-rust.allocs"
+
+  let () =
+    let open Soteria.Stats in
+    let open Soteria.Logs.Printers in
+    register_int_printer ~name:"Allocations" allocs (fun _ -> Fmt.int)
+end
```
N1ark: can we move this to `Rustsymex` with the rest? i'm not even sure this is correct because if you instantiate `Tree_state` several times it will register the printer several times :P

soteria-rust/lib/interp.ml
```diff
@@ -33,70 +34,43 @@ module Make (StateImpl : State.S) = struct
...
-          "Variable %a had value, allocated pointer %a" Expressions.pp_var_id
-            var_id pp_full_ptr ptr];
-        ptr
     | Dead -> error `DeadVariable
+    | Uninit | Value _ ->
+        let* layout = Layout.layout_of binding.ty in
+        if%sat layout.size ==@ Usize.(0s) then
+          (* a ZST has no address; hand out a dangling pointer (à la
+             [NonNull::dangling]) and keep its value in the store *)
+          ok (Sptr.of_address (layout.align :> T.sint Typed.t), Thin)
```
N1ark: i'm ok with this but we need to mention it
```suggestion
          (* a ZST has no address; hand out a dangling pointer (à la
             [NonNull::dangling]) and keep its value in the store.
             This is under-approximating! The exact version would create
             a new nondeterministic address that is divisible by align. *)
          ok (Sptr.of_address (layout.align :> T.sint Typed.t), Thin)
```

giltho:
!run-benchmarks

N1ark (changes requested):
not done

soteria-rust/lib/rust_val.ml
giltho: I was wondering if anything could break this? Like unsafe code with weird transmutes maybe?

soteria-rust/lib/rust_val.ml
giltho: Yeah, thought about it but was lazy lmao

soteria-rust/test/cram/simple.t/run.t
giltho: See how the documentation above says "this is to check the script is working", it's expected to be the wrong number. But I guess I could write the valid number

soteria-rust/lib/interp.ml
```diff
@@ -334,17 +308,25 @@ module Make (StateImpl : State.S) = struct
     | PlaceGlobal g -> resolve_global g
     (* Dereference a pointer *)
     | PlaceProjection (base, Deref) -> (
-        let* ptr = resolve_place base in
-        [%l.debug "Dereferencing ptr %a of %a" pp_full_ptr ptr pp_ty base.ty];
-        let* v = State.load ptr base.ty in
-        match v with
+        (* read the pointer value of [base] without spilling it to the heap *)
```
N1ark, line 317: minor but theres lots of LLM-coded comments id rather not have
```suggestion
```

soteria-rust/lib/interp.ml
N1ark: if you tweak `index_len` to also return the type you can reuse it here

soteria-rust/lib/interp.ml
```diff
@@ -830,17 +972,7 @@ module Make (StateImpl : State.S) = struct
...
-        match place.ty with
-        | TAdt adt when Crate.is_enum adt ->
-            let variants = Crate.as_enum adt in
-            let+ variant_id = State.load_discriminant loc place.ty in
-            let variant = Types.VariantId.nth variants variant_id in
-            Int (BV.of_literal variant.discriminant)
-        (* If a type doesn't have variants, return 0.
-           https://doc.rust-lang.org/std/intrinsics/fn.discriminant_value.html *)
-        | _ -> ok (Int U8.(0s)))
+    | Discriminant place -> load_discr_lazy place
```
N1ark: you've effectively offset a lot of code from the interpreter into the lazy mechanism, and i dont see why? wouldn't  it be enough to just replace these with `resolve_place_lazy` and `load_discriminant_lazy`? overall im not fond of splitting small pieces of behaviour into smaller functions

soteria-rust/lib/interp.ml
```diff
@@ -437,16 +387,208 @@ module Make (StateImpl : State.S) = struct
...
+        | Heap ptr -> from_heap ~variants ptr
+        | Store sp -> (
+            let* store = get_env () in
+            match Store.try_load_discriminant sp store with
+            | Some (DVariant discr) -> ok (Int discr)
+            | Some DUninit ->
+                let* layout = Layout.layout_of sp.ty in
+                if%sat layout.size ==@ Usize.(0s) then
+                  (* There must be a unique variant *)
+                  ok (of_variant (List.hd variants))
```
N1ark: wrong, `enum { B(!), A }` is a ZST

N1ark (changes requested):
finished reading through, confirms my theory i dont like the way its implemented as is -- the changes added even more functions to the massive `let rec ... and ...` that is the interpreter, and there are a lot of small functions that are quite repetitive and seem easy to get wrong; overall i really want to be able to handle the store optimisation far away from any semantic concerns, and this PR clearly mixes the two in a way that is hard to reason about :( ill clone this locally and try cleaning it up

im also not happy about `zst_value`, i actually had done quite a lot to get rid of it and having it come back feels like a failure (even if its much simpler now), ill see if it can be avoided too.

the cleanups and changes may have a small performance cost, but it will improve maintainability is my hope. ill do it and we'll see

soteria-rust/lib/rust_val.ml
N1ark: no a transmute still does a typed read at the end so types match; the cases where it wouldnt is if you transmute or cast a raw pointer, but for those cases the store optimisation doesn't (and should never!) apply, so it's fine

giltho:
Yeah I was unhappy about the mutual recursion but then I realised that, for instance, the store optim needs to be able to evaluated indexes to understand if they're constants...

I already cut the code by half and moved as much logic as I could to store.ml, but not sure how I can do better.

In any case, feel free to improve this as much as you'd like

soteria-rust/lib/interp.ml
giltho: Damn, but then how do I know which one? :o

giltho: I guess I must find the only possible active variant?

N1ark: that's what the decoder does .. but also if it's truly a ZST then calling State.discriminant_of won't perform any reads

soteria-rust/test/cram/simple.t/run.t
N1ark: yeah i think it would make more sense!

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks one last time

soteria-rust/lib/interp.ml
```diff
@@ -437,16 +376,112 @@ module Make (StateImpl : State.S) = struct
...
+        let* len, _ = len_of_indexable ~meta:Thin ~pointee:base.ty in
+        let idx = if from_end then len -!@ idx else idx in
+        (* only a concrete index can be navigated within a value; a symbolic
+           index decays to a heap location *)
+        let*^ z = BV.to_z idx in
+        let* () =
+          assert_ (Usize.(0s) <=$@ idx &&@ (idx <$@ len)) `OutOfBounds
+        in
+        ok Store.Place.{ kind = Index (base_sp, Z.to_int z); origin }
+    | _ -> none ()
```
N1ark, line 451: i'm still not extatic about this, it's a bunch of duplication, but i've accepted my fate.

soteria-rust/lib/interp.ml
```diff
@@ -830,17 +864,27 @@ module Make (StateImpl : State.S) = struct
...
+            ~store:(fun sp store ->
+              let*^ v = Store.try_load sp store in
+              match v with
+              | Value (Enum (discr, _)) -> ok (Int discr)
+              | Uninit -> (
+                  let* dangling = Sptr.dangling_if_zst sp.origin.ty in
+                  match dangling with
+                  | Some d ->
+                      OptionM.lift @@ State.load_discriminant (d, Thin) place.ty
+                  | None -> error `UninitializedMemoryAccess)
```
N1ark, line 896: i'm also not _sure_ this is right... supposedly for `enum Foo { A(u32) }`, you can know the discriminant of the thing without a memory access, so ``​`UninitializedMemoryAccess`` is maybe wrong? but at the same time there's no test for this in the Miri suite, and I'm not even sure the compiler will allow generating a `Discriminant` for an uninitialised place, so I'm going to assume this is ok for now.

soteria-rust/lib/store.ml
```diff
@@ -39,12 +115,54 @@ let declare_value sym value t = declare sym (Value value) t
...
+         match Rust_val.flatten v with
+         | [ Ptr (_, Thin) ] -> Value Rust_val.unit_
+         | [ Ptr (_, Len len) ] -> Value (Int len)
+         | [ Ptr (_, VTable vt) ] ->
+             if
+               Option.is_some_and
+                 (Crate.adt_has_lang_item "dyn_metadata")
+                 (ty_as_adt_opt place.origin.ty)
+             then Value (Tuple [ Tuple [ Ptr (vt, Thin) ]; Tuple [] ])
+             else Value (Ptr (vt, Thin))
```
N1ark, line 156: this will also eventually break with the Big Refactor but I'll handle it there :)

N1ark (approved):
@giltho have a look but I'm happy with how this looks now

giltho:
Happy with this

PR #375, opened 2026-06-10
Add `--offline` to Soteria Rust

N1ark:
Minor; before I forget again about it

### DIFF
--- a/soteria-rust/lib/config.ml
+++ b/soteria-rust/lib/config.ml
@@ -92,6 +92,10 @@ type t = {
       [@docs Sections.frontend] [@names [ "sysroot" ]] [@env "RUST_SYSROOT"]
       (** The sysroot to use for compilation. If not provided, the default
           sysroot is used. *)
+  offline : bool;
+      [@docs Sections.frontend] [@make.default false] [@names [ "offline" ]]
+      (** Whether to compile without accessing the network, which can be useful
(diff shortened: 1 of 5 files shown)


soteria-rust/lib/config.ml
```diff
@@ -92,6 +92,10 @@ type t = {
       [@docs Sections.frontend] [@names [ "sysroot" ]] [@env "RUST_SYSROOT"]
       (** The sysroot to use for compilation. If not provided, the default
           sysroot is used. *)
+  offline : bool;
+      [@docs Sections.frontend] [@make.default false] [@names [ "offline" ]]
+      (** Whether to compile without accessing the network, which can be useful
+          for reproducibility. This will pass --offline to Cargo.. *)
```
giltho, line 101: No env var?

giltho (approved):

giltho:
Should we enable offline in tests?

N1ark:
> Should we enable offline in tests?

we could, it would require doing a `soteria-rust build-plugins` as a setup step but why not? i dont see what the advantage is either though

N1ark:
eh why not sure

soteria-rust/lib/config.ml
N1ark: added one :)

PR #380, opened 2026-06-11
Rust: Disjoint exposed addresses

giltho:
Our remaining false positive in bi-abduction is testing for sentinel values. Libraries like Hashtbl check if a table is empty by checking that its pointer is equal to `Hashtbl::NEW`, where NEW is statically allocated.

Until now, DecayMap would create fresh addresses but not impose any restrictions on them. Now, we ensure that all created base addresses are distinct. We could do more by asserting that none of them overlap, but I'm a bit worried about the performance.


giltho:
!run-benchmarks

N1ark:
anything you can do about the +9.4% or are we just tanking it? i don't mind

soteria-rust/test/cram/perf.t/run.t
```diff
@@ -8,23 +8,13 @@
...
-        (0b00 == extract[0-1](V|12|)) /\ (0b000 == extract[0-2](V|13|))
-  PC 2: (V|3| <u V|2|) /\ (V|4| <u V|2|) /\ (V|4| <u V|3|) /\ (V|5| <u V|2|) /\
-        (V|5| <u V|3|) /\ (V|4| <=u V|5|) /\ (V|4| != V|5|) /\
+  PC 1: (V|3| <u V|2|) /\ Distinct(V|1|, V|6|) /\ (V|4| <u V|2|) /\
+        (V|4| <u V|3|) /\ Distinct(V|1|, V|6|, V|7|) /\ (V|5| <u V|2|) /\
+        (V|5| <u V|3|) /\ (V|5| <u V|4|) /\ Distinct(V|1|, V|6|, V|7|, V|8|) /\
+        Distinct(V|1|, V|6|, V|7|, V|8|, V|9|) /\ Distinct(V|1|, V|6|, V|7|,
+       V|8|, V|9|, V|10|) /\ Distinct(V|1|, V|6|, V|7|, V|8|, V|9|, V|10|,
+       V|11|) /\ Distinct(V|1|, V|6|, V|7|, V|8|, V|9|, V|10|, V|11|, V|12|) /\
+        Distinct(V|1|, V|6|, V|7|, V|8|, V|9|, V|10|, V|11|, V|12|, V|13|) /\
```
N1ark, line 17: this is SO going to blow up

N1ark (approved):
i'm ok with the perf cost, but up to you. my hope is that a nice analysis that makes buckets of disjoints will address this.

PR #382, opened 2026-06-12
Don't decay pointer for null checks in niche optims

giltho:
An issue is to do with, e.g. `Option<NonNull<_>>`. When reading the discriminant of the option, the interpreter performs a pointer decay because it must read the discriminant of the Option, which spans the entire NonNull. That's a bit dumb, because that's effectively just a null check, which should be extremely cheap. Talk about zero-cost abstraction.

Here's a fun insight though: it's cheaper to cast an integer to a pointer than the other way around. So, when the niche is the size of a pointer, instead of reading an integer, we read a pointer. And then if we just want to check whether it's == 0, it's a null check. Otherwise, we decay

Note that the win isn't that substantial because without the charon transformation coming in other PRs, pointer decay still happens in another place for the same pointers we're dealing with here, in calls to __rust_alloc. Combined with the charon change, this amounts to ~1/2 of all decays

### DIFF
--- a/soteria-rust/lib/typed.ml
+++ b/soteria-rust/lib/typed.ml
@@ -114,9 +114,7 @@ module BitVec = struct
 
   let max ~signed l r = ite (gt ~signed l r) l r
   let min ~signed l r = ite (lt ~signed l r) l r
-
-  let sure_is_zero v =
-    match to_z v with Some z -> Z.equal z Z.zero | None -> false
+  let sure_is_zero v = Option.is_some_and Z.(equal zero) (to_z v)
 end
 
 module BV = BitVec
--- a/soteria-rust/lib/typed.ml
+++ b/soteria-rust/lib/typed.ml
@@ -114,6 +114,9 @@ module BitVec = struct
 
   let max ~signed l r = ite (gt ~signed l r) l r
   let min ~signed l r = ite (lt ~signed l r) l r
+
+  let sure_is_zero v =
+    match to_z v with Some z -> Z.equal z Z.zero | None -> false
 end
 
 module BV = BitVec
--- a/soteria-rust/lib/value_codec.ml
+++ b/soteria-rust/lib/value_codec.ml
@@ -200,20 +200,47 @@ struct
         Fields_shape.discriminator -> Types.variant_id ParserMonad.t = function
       | Known v -> ok v
       | Branch { offset = tag_ofs; tag_ty; children; fallback } ->
-          let* tag = query (TLiteral tag_ty, offset +!!@ tag_ofs) in
-          let tag = as_base tag_ty tag in
+          (* You know what's cheaper than Pointer->Integer cast?
+             Integer->Pointer cast. If the tag is pointer-sized, we read it as a
+             pointer anyway, and special case null-checking. Allows for
+             optimising away decays happening for e.g. Option<NonNull<T>>. *)
+          let* tag =
+            if size_of_literal_ty tag_ty = Crate.pointer_size () then
+              let+ v = query (unit_ptr, offset +!!@ tag_ofs) in
+              `Ptr (fst (as_ptr v))
+            else
+              let+ v = query (TLiteral tag_ty, offset +!!@ tag_ofs) in
+              `Int (as_base tag_ty v)
+          in
+          let pp_tag ft = function
+            | `Int v -> Typed.ppa ft v
+            | `Ptr p -> Sptr.pp ft p
+          in
           let pp_info ft () =
-            Fmt.pf ft ", read %a: %s at offset %a" Typed.ppa tag
+            Fmt.pf ft ", read %a: %s at offset %a" pp_tag tag
               (Print.literal_type_to_string tag_ty)
               Typed.ppa offset
           in
+          let cond_for from_ to_ =
+            let int_cond tag =
+              if Typed.equal from_ to_ then tag ==@ from_
+              else from_ <=@ tag &&@ (tag <=@ to_)
+            in
+            match tag with
+            | `Int tag -> ok (int_cond tag)
+            | `Ptr ptr ->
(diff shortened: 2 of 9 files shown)


soteria-rust/test/cram/common.sh
```diff
@@ -0,0 +1,14 @@
...
+	soteria-rust exec "$input_file" --stats stats.json
+
+	local decays
+	decays="$(jq '."soteria-rust.decayed_pointers" // 0' stats.json)"
+
+	if [ "$decays" != "$expected" ]; then
+		echo "check_decays: expected '$expected', got '$decays'" >&2
+		return 1
+	fi
+}
```
N1ark: ok this is getting silly, can you just make the script receive a string (e.g. `decayed_pointers`, `allocs`) and compare it to the expected, so we can reuse the script in both cases?

also, im being a bit annoying so feel free to ignore this, but i don't like having the `soteria-rust` invocation _inside_ the script, because this won't work if we need to call it with CLI flags. instead, could all uses of the script be `soteria-rust exec <args> --stats stats.json && check_stat stats.json decayed_pointers 0`

soteria-rust/test/cram/simple.t/run.t
```diff
@@ -634,3 +627,31 @@ Boolean BitOr must not be assumed true; both operands can be false (issue #376).
...
   [1]
+
+  $ check_decays nonnull.rs 0
+  Compiling... done in <time>
+  => Running nonnull::main...
+  note: nonnull::main: done in <time>, ran 1 branch
+  PC 1: Distinct(V|1-2|) /\ (0x0000000000000008 <=u V|1|) /\
+        (V|1| <=u 0x7ffffffffffffff6) /\ (0x0000000000000008 <=u V|2|) /\
+        (V|2| <=u 0x7ffffffffffffff6) /\ (0b000 == extract[0-2](V|1|)) /\
+        (0b000 == extract[0-2](V|2|))
```
N1ark, line 676: this absolutely looks like two decayed pointers ??? something you did with the stat is wrong

soteria-rust/test/cram/simple.t/nonnull.rs
```diff
@@ -0,0 +1,71 @@
...
+    let b = Box::new(67);
+    Some(NonNull::new(Box::into_raw(b)).unwrap())
+}
+
+fn make_nonnull_res() -> RN {
+    let b = Box::new(42);
+    Ok(NonNull::new(Box::into_raw(b)).unwrap())
+}
+
+fn main() {
```
N1ark: can you be consistent and either stick to `#[soteria::test]`, or to a `main` function? it reads weird to have both


```suggestion
#[soteria::test]
fn match_niched_enums() {
```

soteria-rust/lib/typed.ml
```diff
@@ -114,6 +114,9 @@ module BitVec = struct
 
   let max ~signed l r = ite (gt ~signed l r) l r
   let min ~signed l r = ite (lt ~signed l r) l r
+
+  let sure_is_zero v =
+    match to_z v with Some z -> Z.equal z Z.zero | None -> false
```
N1ark: nit (works with #374 merged)

```suggestion
  let sure_is_zero v = Option.is_some_and Z.(equal zero) (to_z v)
```

soteria-rust/lib/value_codec.ml
```diff
@@ -200,20 +200,47 @@ struct
...
       | Known v -> ok v
       | Branch { offset = tag_ofs; tag_ty; children; fallback } ->
-          let* tag = query (TLiteral tag_ty, offset +!!@ tag_ofs) in
-          let tag = as_base tag_ty tag in
+          (* You know what's cheaper than Pointer->Integer cast?
+             Integer->Pointer cast. If the tag is pointer-sized, we read it as a
+             pointer anyway, and special case null-checking. Allows for
+             optimising away decays happening for e.g. Option<NonNull<T>>. *)
+          let* tag =
+            if size_of_literal_ty tag_ty = Crate.pointer_size () then
```
N1ark: here you're catch `u64`s for i believe no reason; afaik the rust compiler will faithfully keep it as usize, so instead could you do this? if it turns out not to work then revert it but id rather not

```suggestion
            if (match tag_ty with TInt Isize | TUInt Usize -> true | _ -> false) then
```

PS: https://github.com/johnyob/ppx-matches/pull/2 will be merged once you renamed a commit :P

soteria-rust/lib/value_codec.ml
```diff
@@ -200,20 +200,47 @@ struct
...
+            match tag with
+            | `Int tag -> ok (int_cond tag)
+            | `Ptr ptr ->
+                if Typed.equal from_ to_ && Typed.BV.sure_is_zero from_ then
+                  (* fast path *)
+                  ok (Sptr.is_null ptr)
+                else
+                  let+ tag = lift (Sptr.decay ptr) in
+                  int_cond tag
+          in
```
N1ark, line 236: nitpick but the branching in the match is hard to parse
```suggestion
          let cond_for from_ to_ =
            let is_single = Typed.equal from_ to_ in
            let int_cond tag =
              if is_single then tag ==@ from_
              else from_ <=@ tag &&@ (tag <=@ to_)
            in
            match tag with
            | `Int tag -> ok (int_cond tag)
            | `Ptr ptr when is_single && Typed.BV.sure_is_zero from_ ->
                ok (Sptr.is_null ptr)
            | `Ptr ptr -> map int_cond @@ lift (Sptr.decay ptr)
          in
```

N1ark (approved):
super minor nitpicks im sorry, i do agree with the PR as a whole so feel free to merge but id appreciate it if you did the tinyyy things for my peace of mind

N1ark (approved):

PR #384, opened 2026-06-12
Don't decay on `check_overlap` for pointers with != provenance

giltho:
Another fast path that avoid decays.

CI won't path because we need to merge a bunch of other stuff before we truly have no decay in btreeset

### DIFF
--- a/soteria-rust/lib/builtins/intrinsics/impl.ml
+++ b/soteria-rust/lib/builtins/intrinsics/impl.ml
@@ -475,16 +475,22 @@ module M (StateM : State.StateM.S) : Intf.M(StateM).Impl = struct
 
   (** [check_overlap name l r size] ensures the pointers [l] and [r] do not
       overlap for a range of size [size]; otherwise errors, with
-      [`StdErr (name ^ " overlapped")]. *)
+      [`StdErr (name ^ " overlapped")].
+
+      We optimise the path where pointers don't have the same provenance to
+      avoid spuriously decaying pointers. *)
   let check_overlap name l r size =
-    let* l_end = Sptr.offset ~signed:false size l in
-    let* r_end = Sptr.offset ~signed:false size r in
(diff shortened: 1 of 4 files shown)


soteria-rust/lib/builtins/intrinsics/impl.ml
```diff
@@ -475,16 +475,22 @@ module M (StateM : State.StateM.S) : Intf.M(StateM).Impl = struct
...
-    let* r_end = Sptr.offset ~signed:false size r in
-    let* dist1 = Sptr.distance l r_end in
-    let* dist2 = Sptr.distance r l_end in
-    let zero = Usize.(0s) in
-    assert_not
-      (Sptr.have_same_provenance l r &&@ (dist1 <$@ zero &&@ (dist2 <$@ zero)))
-      (`StdErr (name ^ " overlapped"))
+    let same_provenance = Sptr.have_same_provenance l r in
+    if%sure not same_provenance then ok ()
+    else
```
N1ark, line 485: super cute! i had a light doubt about how this interacts when one of the pointers has no provenance but in fact is the address of the decayed pointer (see below), but actually it doesn't matter because even if this lets it through using the pointer will fail bc it has no provenance :) so yes this is fine

```rs
let mut x = 11u64;
// this will have provenance
let ptr2 = &mut x as *mut u64;
// let's make sure the provenance is exposed
let _exposed = ptr2 as usize;
// this will have no provenance, bc it's transmuted
let ptr1 = unsafe {
    let addr_unexposed = std::mem::transmute::<&mut u64, usize>(&mut x);
    std::mem::transmute::<usize, *mut u64>(addr_unexposed)
};
// will this detect the overlap? they have distinct provenances...
std::mem::copy_nonoverlapping(ptr1, ptr2, 1);
```

N1ark (approved):
nice

PR #385, opened 2026-06-12
Don't decay on pointer equality if it can be avoided

giltho:
Again, another fast path. If we know two pointers have the same provenance, or if we know they don't, we can immediately return.

This is another use case for your "lazy &&" I think?

### DIFF
--- a/soteria-rust/lib/builtins/core.ml
+++ b/soteria-rust/lib/builtins/core.ml
@@ -140,12 +140,24 @@ module M (StateM : State.StateM.S) = struct
         let+ res = eval_ptr_binop Eq l r in
         BV.not_bool (cast res)
     | Eq, Ptr (l, meta_l), Ptr (r, meta_r) ->
-        let* meta_eq = eval_meta_eq meta_l meta_r in
         (* Pointer comparison just uses the address! See
            https://doc.rust-lang.org/std/ptr/index.html#provenance *)
-        let+ distance = Sptr.distance l r in
-        let ptr_eq = distance ==@ Usize.(0s) in
-        BV.of_bool (meta_eq &&@ ptr_eq)
+        (* We optimise when the pointers surely have (or surely don't have) the
+           same provenance, to avoid spuriously decaying them. *)
+        let same_provenance = Sptr.have_same_provenance l r in
+        if%sure same_provenance then
+          (* Same allocation: comparing the offsets is enough. *)
+          let+ meta_eq = eval_meta_eq meta_l meta_r in
+          BV.of_bool (meta_eq &&@ (Sptr.ofs l ==@ Sptr.ofs r))
+        else if%sure not same_provenance then
+          (* Distinct allocations live at distinct addresses, so the pointers
+             cannot be equal. *)
+          ok (BV.of_bool v_false)
+        else
+          let* meta_eq = eval_meta_eq meta_l meta_r in
+          let+ distance = Sptr.distance l r in
+          let ptr_eq = distance ==@ Usize.(0s) in
+          BV.of_bool (meta_eq &&@ ptr_eq)
     | Eq, Ptr (p, _), Int v | Eq, Int v, Ptr (p, _) ->
         let v = cast_i Usize v in
         if%sat v ==@ Usize.(0s) then ok (BV.of_bool (Sptr.is_null p))
--- a/soteria-rust/lib/builtins/core.ml
+++ b/soteria-rust/lib/builtins/core.ml
@@ -135,6 +135,7 @@ module M (StateM : State.StateM.S) = struct
     | _, _ -> v_false
 
   let rec eval_ptr_binop (bop : Expressions.binop) l r =
+    let null_or_in_bound p = Sptr.is_null p ||@ Sptr.in_bound p in
     match (bop, l, r) with
     | Ne, _, _ ->
         let+ res = eval_ptr_binop Eq l r in
@@ -149,14 +150,19 @@ module M (StateM : State.StateM.S) = struct
           let+ meta_eq = eval_meta_eq meta_l meta_r in
           BV.of_bool (meta_eq &&@ (Sptr.ofs l ==@ Sptr.ofs r))
         else if%sure
-          (not same_provenance) &&@ Sptr.in_bound l &&@ Sptr.in_bound r
+          (not same_provenance) &&@ null_or_in_bound l &&@ null_or_in_bound r
         then
-          (* Fast path: if two pointer have different provenances AND are in
-             bound of their respective allocations, they cannot be equal (since
-             allocations cannot overlap).
+          (* Fast path: case where two pointers have different provenances. If
+             they are both in bound, then they can't compare equal since two
+             distinct allocations cannot overlap. Similarly, if one of the is
+             null, and the other one has a valid provance and is in bound, then
+             they can't compare equal either because an allocation cannot be at
+             address 0.
 
              Note: pointers that have no provenance have size 0, so they are
-             always out of bound.*)
+             always out of bound, so testing equality of a pointer with valid
+             provenance and a non-0 address with no provenance will bypass this
+             path, as it should. *)
           ok (BV.of_bool v_false)
         else
           let* meta_eq = eval_meta_eq meta_l meta_r in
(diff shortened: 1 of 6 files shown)


soteria-rust/lib/builtins/core.ml
```diff
@@ -140,12 +140,24 @@ module M (StateM : State.StateM.S) = struct
...
+           same provenance, to avoid spuriously decaying them. *)
+        let same_provenance = Sptr.have_same_provenance l r in
+        if%sure same_provenance then
+          (* Same allocation: comparing the offsets is enough. *)
+          let+ meta_eq = eval_meta_eq meta_l meta_r in
+          BV.of_bool (meta_eq &&@ (Sptr.ofs l ==@ Sptr.ofs r))
+        else if%sure not same_provenance then
+          (* Distinct allocations live at distinct addresses, so the pointers
+             cannot be equal. *)
+          ok (BV.of_bool v_false)
```
N1ark: although this was fine for `copy_nonoverlapping` this is incorrect for pointer comparison. consider:
```rust
let x = 11u64;
// this will have provenance
let ptr2 = &x as *const u64;
// this will have no provenance: it's transmuted
let ptr1 = unsafe {
    let addr_unexposed = std::mem::transmute::<&u64, usize>(&mut x);
    std::mem::transmute::<usize, *const u64>(addr_unexposed)
};
assert!(ptr1 == ptr2);
```

you simply cannot skip the decay for this case. even if pointers have different, non-null provenance, they can be equal. consider
```rust
let x = 11u8;
let y = 12u8;
// provenance of x
let ptr1 = &x as *const u8;
// provenance of y
let ptr2 = &y as *const u8;
// still has provenance of y, but address of x
let ptr2 = ptr2.wrapping_sub((ptr2 as usize).wrapping_sub(ptr1 as usize));
assert!(ptr1 == ptr2) // diff provenance, same address.
```

you can _maybe_ hack around it iff:
- they have different provenance
- neither offsets of the pointers are symbolic (i.e. they **cannot** contain a decayed address or part of one)
- you are willing to be UX and say "it's impossible for a concrete address to match a decayed address". i'm ok with that, though i would want a big "This is UX" comment explaining it.

soteria-rust/lib/builtins/core.ml
```diff
@@ -140,12 +140,24 @@ module M (StateM : State.StateM.S) = struct
...
-        let+ distance = Sptr.distance l r in
-        let ptr_eq = distance ==@ Usize.(0s) in
-        BV.of_bool (meta_eq &&@ ptr_eq)
+        (* We optimise when the pointers surely have (or surely don't have) the
+           same provenance, to avoid spuriously decaying them. *)
+        let same_provenance = Sptr.have_same_provenance l r in
+        if%sure same_provenance then
+          (* Same allocation: comparing the offsets is enough. *)
+          let+ meta_eq = eval_meta_eq meta_l meta_r in
+          BV.of_bool (meta_eq &&@ (Sptr.ofs l ==@ Sptr.ofs r))
```
N1ark, line 151: this in contrast is ok as a shortcut, simply because decaying the pointers here would amount to `decayed(loc) + ofs_l == decayed(loc) + ofs_r` so you can naturally cancel the decay

soteria-rust/test/cram/simple.t/run.t
```diff
@@ -150,17 +150,19 @@ Check permissive provenance allows int to ptr casts
...
-  PC 1: Distinct(V|1-2|) /\ (0x0000000000000000 != (V|1| - V|2|)) /\
-        Distinct(V|1-3|) /\ (0x0000000000000000 != (V|2| - V|3|)) /\
-        (0x0000000000000001 <=u V|1|) /\ (V|1| <=u 0x7ffffffffffffffd) /\
-        (0x0000000000000001 <=u V|2|) /\ (V|2| <=u 0x7ffffffffffffffd) /\
-        (0x0000000000000001 <=u V|3|) /\ (V|3| <=u 0x7ffffffffffffffd)
+  PC 1: empty
   
+
+
+
```
N1ark: you keep inserting new lines because your IDE tries removing trailing spaces; i personally disabled formatting of `.t` files

N1ark (changes requested):
nope

giltho:
Fixed. Also added your example as a regression test

soteria-rust/lib/sptr.ml
```diff
@@ -183,6 +183,15 @@ module type S = sig
...
   val is_null : t -> sbool Typed.t
 
+  (** The offset of this pointer within its allocation. *)
+  val ofs : t -> [> sint ] Typed.t
+
+  (** Returns a symbolic boolean characterising whether the pointer is in bound
+      to its allocation. *)
+  val in_bound : t -> sbool Typed.t
+
+  (** The size of the allocation this pointer points to. *)
```
N1ark: comment without function ?

soteria-rust/lib/builtins/core.ml
```diff
@@ -140,12 +140,26 @@ module M (StateM : State.StateM.S) = struct
...
-        let ptr_eq = distance ==@ Usize.(0s) in
-        BV.of_bool (meta_eq &&@ ptr_eq)
+        let same_provenance = Sptr.have_same_provenance l r in
+        if%sure same_provenance then
+          (* Fast path: if two pointer have the same provenance, it's enough to
+             compare their offsets *)
+          let+ meta_eq = eval_meta_eq meta_l meta_r in
+          BV.of_bool (meta_eq &&@ (Sptr.ofs l ==@ Sptr.ofs r))
+        else if%sure
+          (not same_provenance) &&@ Sptr.in_bound l &&@ Sptr.in_bound r
```
N1ark: i think here you're missing checking they both have non-0 provenance; they could point to the same address and one doesn't have provenance

N1ark: add a test for 


`transmute::<usize,*const T>(transmute::<&T, usize>(&x)) == &x `

N1ark: importantly, i think we always create pointers with no provenance as having size 1, so they "are in bounds" as long as they aren't moved -- not sure that's needed though maybe the size could be made 0...? but that's a whole other deal

giltho: Can't pretend I had thought about this case, but I added the test and it already passes because you actually give size 0 to null ptrs :)
```ocaml
 let null () =
      {
        ptr = Typed.Ptr.null ();
        tag = None;
        align = Usize.(1s);
        size = Usize.(0s);
      }
```
I'll add a comment though, this is worth documenting there

giltho:
I also just realised, the following test passes in Miri:
```rs
fn main() {
    let x = 11u8;
    // provenance of x
    let ptr1 = &x as *const u8;
    // still has provenance of x, but address 0
    let ptr1 = ptr1.wrapping_sub(ptr1 as usize);
    assert!(ptr1 == std::ptr::null()) // diff provenance, same address.
}
```

What that means, unfortunately, is that your trick of not introducing `ZeroIfNull` but instead checking that `x != Null` cannot work as an optimisation, because we *always* have to decay `x` to check if it is *different* from Null. So we have to make that pass more complex so we can, instead, check that the pointer is *equal* to null....

giltho:
Well, in general, it means that `Sptr.is_null` is OX-unsound 🙃 Null checks should *always* check if the ptr is in bound to its allocation

giltho:
Ok I've improved the fast path with different provenance by saying that if you are comparing a pointer with valid provenance that is in bound to its allocation with a null pointer, then you know they can't be equal. Together with https://github.com/soteria-tools/obol/pull/3, it gets rid of all remaining decays in Btreeset :)

N1ark (approved):

N1ark:
> Well, in general, it means that `Sptr.is_null` is OX-unsound 🙃 Null checks should _always_ check if the ptr is in bound to its allocation

oh noooo yeah that's not good -- created #396 

> we always have to decay x to check if it is different from Null

not true, we can still be efficient with it, while being _less UX_. if a pointer has non-0 provenance, and its offset is entirely concrete, then we can say that the pointer is not null. the UX aspect of this is just "you cannot magically (i.e. concretely) guess a pointer's address", which is Fine™, and will catch the majority of cases. 
furthermore, if the pointer _is_ in fact null, then it has been decayed so decaying it again to check is costless.

let's fix that separately, but thanks for raising it! PR is good as is otherwise

giltho:
Sorry I meant "we always may have too" meaning "we can't trust a non-null location" but yeah
Honestly this is a good enough under-approximation. What mad person would construct a null pointer with a valid provenance 🤣

PR #386, opened 2026-06-12
Use the new Charon builtin called `ZeroIfNull`

giltho:
This will crash until we merge the necessary Charon/Obol changes


soteria-rust/lib/interp.ml
```diff
@@ -583,6 +548,8 @@ module Make (StateImpl : State.S) = struct
     in
     function
     (* Handle builtins separately *)
+    | FnOpRegular { kind = FunId (FBuiltin ZeroIfNull); generics = _ } ->
+        perform_call (Synthetic ZeroIfNull)
```
N1ark, line 552: rejecting early; this will never work and never happen in Charon etc etc; Nadri (and I) actively wants to remove builtin functions (https://github.com/AeneasVerif/charon/issues/1164); they're a hack and bad. id need to look at where do you make this change in charon but ideally this should e.g. be a `UnOp`. the issue though is still getting it merged into Charon, talk with Nadri to see if there's a compromise hes willing to take, because i like the idea of having small builtin operators for optimisations for us but this one won't slide i can tell you that, and i really want to have us use upstream Charon as much as possible

N1ark (changes requested):
see comment; this is unmergeable as is :/

soteria-rust/lib/interp.ml
N1ark: a possibility is also litearlly just translating these comparisons into `Binop (Eq, ptr, ConstantExpr::PtrNoProvenance(0))`; this will be **a lot** more likely to slide. I don't understand why this returns a usize though

N1ark: ah it returns a `usize` because the code it substitutes returned a usize. then you would reconstruct it as this, were `_x` is the original place and `_tmp` is a fresh new `bool` local
```
_tmp := _ptr == PtrNoProvenance(0)
_x := _tmp as usize
```

giltho:
The only change that will be required once https://github.com/soteria-tools/obol/pull/3 is merged will be to bump the obol version and promote the test :)

PR #390, opened 2026-06-12
Support calls to `dyn FnOnce`

### LINKED ISSUE #387: Support calls to trait objects that have `Self` as an argument
Although technically traits with a function receiving `Self` without a pointer [are not dyn-compatible](https://doc.rust-lang.org/reference/items/traits.html#r-items.traits.dyn-compatible.associated-functions), `FnOnce` _is_ dyn-compatible, which causes issues.

We should fix this! Reproducer: `(Box::new(&|| {}) as Box<dyn FnOnce()>)()`

N1ark:
Fixes #387 

idk i sent the slopper at this thinking it would be hard and it's actually so stupid im disappointed i havent found this fix earlier; i had attempted fixing it and it was .. a mess

### DIFF
--- a/soteria-rust/lib/interp.ml
+++ b/soteria-rust/lib/interp.ml
@@ -1073,9 +1073,23 @@ module Make (StateImpl : State.S) = struct
         let* args =
           map_list (List.combine3 args in_tys exp_tys)
             ~f:(fun (arg, from, to_) ->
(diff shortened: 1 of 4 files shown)


giltho (approved):
💥

N1ark:
i want #392 to be merged before merging this :)

PR #391, opened 2026-06-12
Always use a sysroot

### LINKED ISSUE #322: Automate the creation and use of a sysroot
Right now, the process to creating a sysroot is:
```sh
$ obol toolchain-path
~/.rustup/toolchains/nightly-2026-02-07-aarch64-apple-darwin
$ cargo +nightly-2026-02-07 miri setup --print-sysroot
/Users/me/Library/Caches/org.rust-lang.miri
$ RUST_SYSROOT=/Users/me/Library/Caches/org.rust-lang.miri soteria-rust ...
```

It'd be good if there was a way of setting up the sysroot and detecting if it's already there or not, so that all of this is hidden behind `soteria-rust`

N1ark:
Fixes #322 

surprised how simple this was :) nice
there is an (i believe) inevitable perf cost to this; we can look into building our own sysroot that has some optimisations enabled, to see if we can reclaim this difference


N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

ok this should work with a sysroot now

giltho:
I'm afraid we can't do much about the performance, but let's delay merging this until we're sure?

N1ark:
> I'm afraid we can't do much about the performance, but let's delay merging this until we're sure?

i'm not sure what you want to do? unfortunately we absolutely need a sysroot if we want to run real code, im okay with the performance hit. 

also CI is having a weird error for this benchmark so im investigating that first

anyways i want #392 to be merged before merging this :)

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

soteria-rust/test/cram/simple.t/run.t
```diff
@@ -694,6 +701,8 @@ Test we can use ptr::metadata to get the metadata of a trait object; this used t
   PC 1: (0x0000000000000000 == V|1|) /\ (0x0000000000000000 == V|1|)
   PC 2: (0x0000000000000001 <=u V|1|)
   
+  check_stat: expected '0', got '3' for decayed_pointers
+  [1]
```
N1ark, line 705: : ( ill try investigating this but urgh

giltho:
What did we decide here?

N1ark:
> What did we decide here?

Up to you. I think doing this is an improvement as users will eventually reach something thats not in the default sysroot, and then they'll have to do this anyways. I would rather we merge this and work on improving perf where we can, rather than not merging this bc of perf when it is an inevitable thing to use.

I wanted to try building our own sysroot (rather than using Miri's) to see if we could enable some optimisations that dont remove "most" UB (e.g. inlining), to allow getting some of this perf back, but never got around to it and it seems rather incremental so id be happy just doing that later

giltho (approved):
If you want, go ahead

PR #392, opened 2026-06-13
Conformance benchmark in CI

N1ark:
Add our pre-existing Kani and Miri conformance test suite to CI; when benchmarks are run, we also run against the Kani and Miri suite, and check what tests we do and don't pass :) 

It will also output when run-benchmarks is used if there are any test changes.

I also redid the styling of the benchmarks, its nicer now.


N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
we can't really test this until it's merged into main, youll just have to trust that its right i guess :')

giltho (approved):
I mean I obviously didn't read the python scripts but heh

giltho:
One needs to pick the correct version of obol now

PR #394, opened 2026-06-13
Minor Tree Borrows optimisation

N1ark:
Minor thing I spotted while trying to improve Tree Borrows performance further. It doesn't do much but its a slight improvement and it's aesthetically nicer. Also the `.gitignore` was wrong so new Claude skills weren't tracked :P 

Also added two benchmarks, one that has a call nested N deep, which is worst case on the amount of tags to track (since tags can never be discarded), and one that has calls N wide (2^N total), which exercises a hopefully more realistic workload where we only track `O(log(N))` of tags

### DIFF
--- a/.claude/skills/profile-soteria/SKILL.md
+++ b/.claude/skills/profile-soteria/SKILL.md
@@ -215,6 +215,10 @@ measurement trap, or a benchmark that isolates a subsystem well, add it to
 `references/` rather than keeping it in your head.
 
 - `references/ocaml-profiling.md` — concrete CPU / allocation / Z3-boundary
-  profiling for Soteria's OCaml + subprocess architecture.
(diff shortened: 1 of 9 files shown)


giltho (approved):
Not sure I understand 100% what's happening cause I don't know what `With_foreign` is, but that's small enough a change that I can trust it's good

giltho:
!run-benchmarks

N1ark:
> Not sure I understand 100% what's happening cause I don't know what `With_foreign` is, but that's small enough a change that I can trust it's good

it's basically just a functor to create a module that allows folding/iterating over two maps with the same key type at the same time; this means the second map lookup we were doing in the loop is done automatically as part of the iteration, rather than separately, which is slightly better

PR #397, opened 2026-06-14
`not_impl` receives an fmt instead of a string

giltho:
As in the title. Something that's been annoying me for a while but I didn't know the incantation to put in the signature in the mli. Claude found it


N1ark:
omgggg i had the exact same thought i just didnt know how to do the right thing so i never bothered 🫠 

kind of a shame when lifting you have to resort to `Fmt.kstr` but i guess it's ok

N1ark (approved):

giltho:
> kind of a shame when lifting you have to resort to Fmt.kstr but i guess it's ok

Yeah, there's no way of doing better. I challenged Claude on this specifically but OCaml is what it is...

PR #400, opened 2026-06-14
Minor fixes and cleaning up

giltho:
PR number 400 🎉

- Fixes for creating a bound when checking if a pointer is dangling
- Add debugging logs for `Tree_state`, it was very opaque without that when reading logs
- When producing a bound, properly produce the fact that the bound must be further than the max offset of the object
- Fix that hack that declares missings as UBDangling to only be triggered in whole program mode


N1ark:
!run-benchmarks

N1ark (approved):

N1ark:
ok when the benchmark is done :) but its minor actually i dont think this changes any behaviour anyways

PR #401, opened 2026-06-14
Fix Z3 leaks

giltho:
In my script that run Soteria with a timeout, I noticed that z3 subprocesses would be left alive after execution.
This implements an abstraction for killing all subprocess when receiving sigkill/sigterm signals.

I'm not an expert at what's happening in there, so I'm not 100% sure if this is correct...


giltho:
!run-benchmarks

soteria/lib/soteria_std/subprocess.ml
```diff
@@ -0,0 +1,91 @@
...
+    Sys.set_signal signum Sys.Signal_default;
+    Unix.kill (Unix.getpid ()) signum
+  in
+  let install signum =
+    match Sys.signal signum (Sys.Signal_handle handle) with
+    | Sys.Signal_default -> ()
+    | prev -> Sys.set_signal signum prev
+  in
+  install Sys.sigint;
+  install Sys.sigterm
```
N1ark, line 74: this is the only part im just trusting you to be right, the rest is ok

soteria/lib/soteria_std/sys.ml
```diff
@@ -6,3 +6,8 @@ let with_working_dir dir f =
   let old_dir = getcwd () in
   chdir dir;
   Fun.protect ~finally:(fun () -> chdir old_dir) f
+
+(* When used in Unix.kill, the "null signal" delivers nothing but still runs
+   kill(2)'s error checks, so it reports whether a pid still names a process --
+   the POSIX liveness probe. *)
+let signull = 0
```
N1ark: this documentation isn't super clear but whatever sure
```suggestion
(** When used in Unix.kill, the "null signal" delivers nothing but still runs
   kill(2)'s error checks, so it reports whether a pid still names a process --
   the POSIX liveness probe. *)
let signull = 0
```

soteria/lib/dune
```diff
@@ -14,7 +14,6 @@
   patricia-tree
   printbox-text
   progress
-  soteria.smt
```
N1ark, line 17: :( this means people can't depend on `soteria.smt` without bringing all of `soteria` in right

N1ark (changes requested):
the loss of `soteria.smt` as a standalone module makes me incredibly sad; is it impossible to keep? honestly i'd be happy to create a separate module for `soteria_std`, so that then we have `soteria.smt` which just depends on `soteria.soteria_std`, and `soteria` which depends on both.

giltho:
Ok but question, why? 

Importing soteria imports soteria.smt *anyway* because it imports bv_values. So there's no value in keeping it separate

N1ark:
> Ok but question, why?
> 
> Importing soteria imports soteria.smt _anyway_ because it imports bv_values. So there's no value in keeping it separate

the other way around; soteria.smt doesn't import soteria (if im not misunderstanding), which is nice if people were to want a better replacement for simple smt

giltho:
I don't think soteria.smt has so much value on its own really. Bv_value does

N1ark (approved):
okay sure :( rest in peace `soteria.smt`, i will miss you

PR #403, opened 2026-06-15
Support more atomics, some externs, fix benchmark

### LINKED ISSUE #388: Fix plugins breaking when compilation settings change
We handle linking our plugins rather hackily; this means if a user changes compilation settings that affect plugins between runs, they get faced with a rather ugly error that is fixed by running `soteria-rust build-plugins`. 

We should either:
- smart solution: detect when plugins need to be cleaned, and rebuild them fresh
- easy solution: catch this error, and force a clean + rebuild of the plugins
- ideal solution: fix how we link the libraries we build so that we specify the exact hash of the dependencies, rather than letting rustc figure out what the library is (i don't know if this is feasible).

Reproducer (one of many):
```
$ soteria-rust exec my_file.rs
<works>

$ soteria-rust exec my_file.rs --sysroot my_sysroot
Compiling... errored
error: Compilation error:
error[E0464]: multiple candidates for `rmeta` dependency `soteria` found
  --> /Users/opale/Documents/GitHub/soteria-agent/./soteria-rust/test/cram/simple.t/fn_ptr.rs:29:3
   |
29 | #[soteria::test]
   |   ^^^^^^^
   |
   = note: candidate #1: /Users/opale/Documents/GitHub/soteria-agent/_build/install/default/share/soteria-rust/plugins/soteria/target/aarch64-apple-darwin/debug/deps/libsoteria-37759632a75d1b54.rmeta
   = note: candidate #2: /Users/opale/Documents/GitHub/soteria-agent/_build/install/default/share/soteria-rust/plugins/soteria/target/aarch64-apple-darwin/debug/deps/libsoteria-ac60e32f5d4316b8.rmeta

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0464`.
ERROR Code failed to compile
```

N1ark:
Minor improvements extracted from #391 while we figure that out
- Ensure plugin building doesn't break when compilation settings change -- closes #388 
- Fix a couple stubs (`_tlv_atexit`, `_env`) that were incorrectly handled with function stubs, when they're externs
- support `std::ptr::Pointee` with polymorphism
- add the missing externs to suppport thread locals on Linux: `__cxa_thread_atexit_impl`, `__dso_handle`, `pthread_key_create`, `pthread_setspecific`, `pthread_key_delete` in system + `malloc` and `free` in libc
- fix a potential soundness bug where `Sptr` used checked operators even when overflow wasn't actually checked
- fix the first test of our benchmark never being run lmao

### DIFF
--- a/soteria-rust/test/cram/simple.t/run.t
+++ b/soteria-rust/test/cram/simple.t/run.t
@@ -636,7 +636,7 @@ Boolean BitOr must not be assumed true; both operands can be false (issue #376).
   [1]
 
 Test that allocating a box only requires two heap allocation (thanks to the store optimisation): one for the contents of the box, and one for the box that we pass to the drop glue.
-FIXME: because of the sysroot, there's a bunch of extra allocations from added function calls... i'm not sure we can avoid them all.
+FIXME: now that named consts are globals, there is in fact a third allocation: the one for <i32 as SizedTypeProperties>::LAYOUT. We should extend the store optimisation to handle globals; in particular we have a guarantee they can't be written to, so it's likely the optimisation will perform really well.
   $ soteria-rust exec box.rs --stats stats.json && check_stat stats.json allocs 2
   Compiling... done in <time>
   => Running box::main...
--- a/soteria-rust/test/cram/simple.t/run.t
+++ b/soteria-rust/test/cram/simple.t/run.t
@@ -4,24 +4,18 @@ Test memory leaks
   => Running leak::main...
   error: leak::main: found issues in <time>, errors in 1 branch (out of 1)
   warning: Memory leak in leak::main
-      --> $RUSTLIB/library/alloc/src/alloc.rs:101:9
-  101 |            __rust_alloc(layout.size(), layout.alignment())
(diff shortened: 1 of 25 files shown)


soteria-rust/test/cram/simple.t/run.t
```diff
@@ -636,7 +636,7 @@ Boolean BitOr must not be assumed true; both operands can be false (issue #376).
   [1]
 
 Test that allocating a box only requires two heap allocation (thanks to the store optimisation): one for the contents of the box, and one for the box that we pass to the drop glue.
-FIXME: now that named consts are globals, there is in fact a third allocation: the one for <i32 as SizedTypeProperties>::LAYOUT. We should extend the store optimisation to handle globals; in particular we have a guarantee they can't be written to, so it's likely the optimisation will perform really well.
+FIXME: because of the sysroot, there's a bunch of extra allocations from added function calls... i'm not sure we can avoid them all.
```
giltho: Revert this?

giltho (approved):

PR #405, opened 2026-06-15
More efficient Logging

giltho:
## Improvement

So here's a fun one. The old logging interface used to take the form:
```ml
L.debug (fun m ->m "%a" pp x)
```
Why would we allocate a closure? One could have just had `L.debug "%a" pp x` and we could make it work (just like `Fmt.failwith` or, currently, `L.failwith`.

Well, there is a perf improvement in the case where you *don't* want to log. I originally implemented this following the results given by this benchmark: https://github.com/mirage/ocaml-git/pull/130#issuecomment-149326614, which noted that keeping everything in a closure would make it faster in the case where the closure isn't applied.

**But** it still requires allocating a closure, even when we don't care about the logging at all! The only way to avoid this is to manually guard on the log level for callsite:
```
if should_log Debug then L.log "%a" pp x;
```

Obviously this is unwieldy and you don't want the user to be typing all of that. **Except** we now have a ppx for our logging. So really we can just change the way it desugars to do the right thing. And that's what this PR does

## Micro benchmarks

When not logging, the cost is no reduced:
- from 4 allocated words to 0 allocated words, reducing GC pressure on the minor heap
- by 40% in cpu clock

When logging is enabled, nothing changes.

## Will it change anything?

No. It's just cute.

That being said, if we ever parallelise Soteria, we might want to have a separate thread that is in charge exclusively of logging. This means we will have to send data across threads, even if no logging is performed. This transformation means that we, in fact, will never do this if we don't need to log, freeing up the logging thread.

### DIFF
--- a/soteria/bench/dune
+++ b/soteria/bench/dune
@@ -1,0 +1,5 @@
+(executable
+ (name logs_bench)
+ (libraries soteria unix)
+ (preprocess
+  (pps ppx_symex)))
--- a/soteria/bench/logs_bench.ml
+++ b/soteria/bench/logs_bench.ml
@@ -1,0 +1,37 @@
+open Soteria.Logs.Import
+
+let iters = 50_000_000
+
+let bench name f =
+  Gc.full_major ();
+  let w0 = Gc.minor_words () in
+  let t0 = Unix.gettimeofday () in
+  f ();
+  let t1 = Unix.gettimeofday () in
+  let w1 = Gc.minor_words () in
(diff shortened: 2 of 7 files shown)


giltho:
!run-benchmarks

giltho:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark (approved):
unconvinced about the performance improvement but i like the idea so ok

PR #406, opened 2026-06-16
Fixes for `HashTable`

N1ark:
Some fixes to get us closer to support `HashTable`/`hashbrown` in general
don't be scared, most of the diff is generated code for SIMD intrinsics.

- Support some subset of SIMD; hashbrown compiled on most processors will use them, and they're actually very simple to implement if we ignore all the corner cases we don't care about yet ^^'
- Stub `std::hash::BuildHasher::hash_one` to always hash to `0`, because being symbolic in hashes leads to path explosion with hashbrown, as it uses the lower N bits of the hash to pick buckets. 
- Also modify `hashmap_random_keys` for the same reason; being symbolic in our hashes leads to path explosion. Furthermore, we can't easily randomise the output of this function because of branching (and id want to avoid adding state to Rustsymex for something so irrelevant)
- Fix constant refs to globals being interpreted as creating new allocations, which led to weird stuff like `assert!(const { &FOO as *const _} == const { &FOO as *const _})` failing


soteria-rust/lib/builtins/optim/impl.ml
```diff
@@ -117,6 +117,14 @@ module M (StateM : State.StateM.S) : Intf.M(StateM).S = struct
   let begin_panic ~m:_ ~msg =
     match msg with Ptr msg -> do_panic ~msg () | _ -> do_panic ()
 
+  (* ---- hashing ---- *)
+
+  (* Replace the real [SipHasher] with the *constant* hash: every value hashes
+     to 0. This is a valid hash function, and avoid branch explosion from
+     symbolic hashes, as e.g. hashbrown uses the lower bits of the hash to pick
+     a bucket. *)
+  let hash_one ~types:_ ~t_self:_ ~t:_ ~self:_ ~x:_ = ok (Typed.BitVec.u64i 0)
```
giltho:
```ocaml
if Soteria.Symex.Approx.As_ctx.is_ox () then warn_once ...
```

soteria-rust/lib/builtins/system/impl.ml
```diff
@@ -5,8 +5,13 @@ module M (StateM : State.StateM.S) : Intf.M(StateM).S = struct
...
 
-  let hashmap_random_keys ~(fun_sig : Types.fun_sig) =
-    Encoder.nondet_valid fun_sig.output
+  (* Returning *symbolic* keys makes every hash symbolic, causing branch
+     explosion; we instead stick to a concrete hash. *)
+  let hashmap_random_keys ~fun_sig:_ =
+    let k0 = 0x0123456789abcdefL in
+    let k1 = 0x94D049BB133111EBL in
+    let to_u64 i = Int (Typed.BV.u64 (Z.of_int64 i)) in
+    ok (Tuple [ to_u64 k0; to_u64 k1 ])
```
giltho, line 22: Same, `if is_ox ...`
And can we at least generate actually random values starting from a constant seed?

giltho (approved):

soteria-rust/lib/builtins/system/impl.ml
N1ark: how do you do that with branching ?

N1ark: like you branch, backtrack, run again and get a different value? that's odd

PR #413, opened 2026-06-18
Use GADT for bv_value

giltho:
We're not merging this, I was just very curious to see what it would look like, and if there's any insight we can take away :)


soteria/lib/bv_values/svalue.ml
```diff
@@ -196,74 +336,125 @@ module Binop = struct
...
+  | UnPtr : 'o Unop.on_ptr * sptr t -> 'o t_kind
+  | BvArith : Binop.arith * sbv t * sbv t -> sbv t_kind
+  | BvCmp : Binop.cmp * sbv t * sbv t -> sbool t_kind
+  | FArith : Binop.farith * sfloat t * sfloat t -> sfloat t_kind
+  | FCmp : Binop.fcmp * sfloat t * sfloat t -> sbool t_kind
+  | BoolBin : Binop.boolean * sbool t * sbool t -> sbool t_kind
+  | Eq : 'a t * 'a t -> sbool t_kind
+  | Nop : Nop.t * 'a t list -> sbool t_kind
+  | Ite : sbool t * 'a t * 'a t -> 'a t_kind
+  | Exists : (Var.t * packed_ty) list * sbool t -> sbool t_kind
```
N1ark, line 401: hm this is interesting, that it splits everything by the typing, i kinda like it

N1ark (commented):
pretty fun; if you fix CI id be curious to run the benchmarks on this
i quite dislike the changes done in `Symex.Value`, i think it makes for a less pleasant interface than what we have; 
if i understand correctly, this give a better confidence on the fact the ghost type matches the value then?

giltho:
!run-benchmarks

giltho:
!run-benchmarks

PR #416, opened 2026-06-18
Allocation-free hashing for Bv_values

giltho:
I let Claude explore while I was working on something else.
It's a good insight, the shitty hashing function I had designed before was allocating 🤦‍♂️

### DIFF
--- a/soteria/lib/bv_values/svalue.ml
+++ b/soteria/lib/bv_values/svalue.ml
@@ -335,19 +335,41 @@ module Hcons = Hc.Make (struct
 
   let equal = equal_t_node
 
-  (* We could do a lot more efficient in terms of hashing probably, if this ever
-     becomes a bottleneck. *)
+  (* Allocation-free structural hash *)
+  let[@inline] combine h x = (h * 65599) + x
+
+  let rec hash_ty = function
+    | TBool -> 1
+    | TFloat p -> combine 2 (FloatPrecision.size p)
+    | TLoc n -> combine 3 n
+    | TPointer n -> combine 4 n
+    | TSeq ty -> combine 5 (hash_ty ty)
+    | TBitVector n -> combine 6 n
+
   let hash { kind; ty } =
-    let hty = Hashtbl.hash ty in
+    let h = hash_ty ty in
     match kind with
-    | Var _ | Bool _ | Float _ | BitVec _ -> Hashtbl.hash (kind, hty)
-    | Ptr (l, r) -> Hashtbl.hash (l.tag, r.tag, hty)
-    | Seq l -> Hashtbl.hash (List.map (fun sv -> sv.tag) l, hty)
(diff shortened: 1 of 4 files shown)


N1ark:
!run-benchmarks

N1ark (approved):
i'm ok with this; ideally we should just use https://github.com/sim642/ppx_deriving_hash though; i think if you define `let hash (x: t) = x.tag` you can then derive the hash on `t_node` and `ty` and it would produce the same code with significantly less maintenance burden

N1ark:
if you want to merge this as is i'm okay with it, if you want to use the deriving i'd prefer it, but i dont mind

giltho:
So, I did look at ppx hash when I created Soteria. Only existing library at the time was JaneStreet's ppx_hash
image

Note how the description of ppx_deriving_hash is this 😂
image

Anyway, the hash function that is in this PR is tailored for the Hashmaps and structure so I'd say we keep this?

N1ark:
i dont think i understand your comment; i mean that library now exists, so why not use it? idk what you mean by 

> the hash function that is in this PR is tailored for the Hashmaps and structure

the derived hash would be .. basically exactly the same except with a different multiplier ?

giltho:
Hmm, ok I'll stop being lazy. I just wanted to check the code it produces before I merge it

giltho:
As I feared, the `ppx_deriving_hash` is written so that allocation is performed when folding over lists (in Nop).
That being said, we can use it to derive hash for everything else than t_kind, because we're currently using Hashtbl.hash for the operators for instance, and that's costly! We get ~20% faster than by using a combination of:
- manual for t_kind
- deriving hash for everything

N1ark:
oh interesting, can you share the generated code? and yeah then go for that ig; when you say 20% faster it's over hashing arbitrary values right? still nice

soteria/lib/bv_values/svalue.ml
```diff
@@ -335,19 +335,41 @@ module Hcons = Hc.Make (struct
...
+  (* Allocation-free structural hash *)
+  let[@inline] combine h x = (h * 65599) + x
+
+  let rec hash_ty = function
+    | TBool -> 1
+    | TFloat p -> combine 2 (FloatPrecision.size p)
+    | TLoc n -> combine 3 n
+    | TPointer n -> combine 4 n
+    | TSeq ty -> combine 5 (hash_ty ty)
+    | TBitVector n -> combine 6 n
```
N1ark, line 347: can't you also derive hash for ty?

giltho: I looked again sorry, the issue is not just for lists, but for any recursive type. I'll PR ppx_derivin_hash to fix this though

N1ark: let's merge this as is for now and bump when the fix has been merged then :)

N1ark:
!run-benchmarks

PR #418, opened 2026-06-19
[DO NOT MERGE] Make Rust values a kind of typed value

N1ark:
Opening as draft; I will try splitting this in 2-3 chunks to keep it reviewable and easier to understand

# The Migration

## `Bv_values` as a functor

Make `Bv_values.Svalue`, and all other modules there, functors on an _extension module_, which for now is defined as follows. This is the minimal shell required for `Svalue` to implement all its functions, along with `Expr` and `Encoder`.
```ocaml
module type Value_ext = sig
  type t [@@deriving eq, ord]
  type ty [@@deriving eq, ord]
  
  type super_t := (t, ty) Svalue.t
  type super_ty := ty Svalue.ty

  val pp : super_t Fmt.t -> t Fmt.t
  val pp_ty : ty Fmt.t
  val iter_vars : (super_t -> unit) -> t -> unit
  val hash : t -> int
  val mk : super_ty -> t -> super_t
  val eval : (super_t -> super_t) -> t -> t

  val apply_subst :
    (missing_var:(Var.t -> super_ty -> super_t) ->
    'subst ->
    super_t ->
    super_t * 'subst) ->
    missing_var:(Var.t -> super_ty -> super_t) ->
    'subst ->
    t ->
    t * 'subst

  val encode_ty : (super_ty -> Soteria_smt.sexp) -> ty -> Soteria_smt.sexp
  val encode_value : (super_t -> Soteria_smt.sexp) -> t -> Soteria_smt.sexp
end
```

Next, made `Typed` a functor, that receives an extension and takes care of instantiating `Svalue` & co, packaging them all neatly together. This matters because **instantiating `Svalue.Make` creates a new `Hcons` module**; two `Svalue` instantiations with the same extension are not interchangeable, and passing values from one to another will cause errors, as the `Hcons` tags won't line up.

This change is mostly churny, the behaviour is very much the same. We also expose `Dummy_ext` with `type t = |`.

## Rust values as an extension

Next, defined an extension for Rust values, to replace `rust_val`; it defines the type of values and types as below. This is an almost 1-1 translation of Rust values, with the addition of typing.
```ocaml
type ty =
  | TAdt of Types.type_decl_ref
      (** invariant: the type decl ref must be that of an {b enum or union};
          structs and tuples go through [TTuple] *)
  | TTuple of svty list
  | TThinPtr
  | TFullPtr
  | TPolyType

type ptr = { ptr : sv; tag : Ptr_tag.t option; size : sv; align : sv }

type t =
  | Ptr of sv * sv option  (** pointer, with optional meta *)
  | ThinPtr of ptr
      (** thin pointer, without metadata but with extra info on the pointer *)
  | Enum of sv * sv list  (** discriminant * values *)
  | Tuple of sv list  (** contains ordered values *)
  | Union of (sv * sv * sv) list
      (** list of blocks in the union, with their offset and size *)
  | PolyVal of Charon.Types.type_var_id
      (** The opaque value of a type variable, identified by (type variable
          index, unique identifier). *)
```

Importantly, we distinguish structs and tuples from enums and unions when typing them; we type structs as a simple tuple type which stores a list of inner types, because resolving the fields of a struct would require recursively traversing the struct, which I wanted to avoid. For enums (and probably unions) this will however be unavoidable, since we need to know the variants; possibly later we may re-define enums as a `ty list list`, where `ty list` is a single variant?

Note **we do not implement the encoding for Rust values yet**. This PR does not change behaviour, and as such Rust values never get sent to the solver, as before.

## Typed Rust values

We then define a `typed.ml` and `typed.mli` pair of files to expose additional typed functions over these new values, to end up with a similar behaviour to what `rust_val` used to have. **I have not found a way to let `typed.ml` know that `'a Typed.t = Svalue.t` without it collapsing the typing across all of Soteria Rust**; this makes `typed.ml` very unwieldy, with lots of `type_` and `untyped` everywhere. I would really want to improve this, but haven't found a way yet; I'm working on this.

Similarly to above, **we do not implement all field accessors for Rust values**, as for now Rust symbolic values can never be generated. We instead just pattern match on the structure, and fail with a `TODO` if that branch is reachable (which it shouldn't be). 

## Porting Soteria Rust

Having typed values (and thus possibly symbolic variables for Rust values) means we **cannot** pattern match over Rust values anymore! This is both a bit annoying, but also very nice because it forces us to instead rely on Rust types everywhere, which is a lot more thorough and trustworthy. This basically enforces the fact that Rust values are _always_ well-typed, wrt their type in the interpreter!

I've also worked a bit on getting rid of most uses of `Typed.cast` ; there are only **4** left and they are all around annoying higher-kinded functions where changing them leads to annoying type errors around the polymorphic variant.

There are some implicit casts done by `match%ty`, a new PPX for pattern matching over a value's _type_, however that's always safe to do and means we are only doing casts that we know are correct, so all is good!

#  What's next

## Split

I will first work on splitting this into reviewable chunks, in separate PRs (except if you would prefer all of this to be a single PR, with structured commits? i think PRs are nicer as they force CI to pass at each step)
The structure I'm aiming for is:

1. make `Bv_values` a functor, use the dummy extension in C and Rust
2. rework some of Soteria Rust to avoid pattern matching on Rust values, do the bulk of the structural changes (e.g. new functions, remove functions, etc.)
3. hide the constructors from `Rust_val`, force all existing code to go through functions
4. Make the Rust extension, hopefully by then the changes to do in the interpreter are minor and mostly about replacing `rust_val` with ~`Typed.(T.any t)`

## `'a Typed.t = Svalue.t`

Find a way to make the `typed.ml` file in Soteria Rust know that typed values are equal to untyped ones, to get rid of a lot of boilerplate (or worse, `List.map typed vs`, which is an actual performance problem).

## Fix up touches 

There is a list of todos, in `TODOS.md`; I'll try going through them in parallel and seeing what can be done to improve the ergonomics of it all.


giltho:
Clicked send to early, I'm still going through and typing

Nitpicking but `super_t` is not a great name. Possibly `ext_t`?

> This matters because instantiating Svalue.Make creates a new Hcons module; two Svalue instantiations with the same extension are not interchangeable, and passing values from one to another will cause errors, as the Hcons tags won't line up.

Very interesting and important. Can we ensure statically somehow that we're not making this mistake?

giltho:
Hmm my conclusion is that I'd like a better vision on where this is going before we merge anything :/

N1ark:
> Nitpicking but `super_t` is not a great name. Possibly `ext_t`?

it's not a real type, it uses `:=`, it's just for brevity in the rest of the signature. also no, because `ext_t` makes it sound like its the extension's type; `super_t` is the type of the Svalue with the extension applied to it. but regardless; it gets erased, it's just for brevity

N1ark:
> Hmm my conclusion is that I'd like a better vision on where this is going before we merge anything :/

since i opened the PR, i fixed typed.ml to avoid all the ugly `type_` etc wrappers, so that flows a lot more nicely; what do you dislike?

> Very interesting and important. Can we ensure statically somehow that we're not making this mistake?

hmm, not sure? possibly if we make `Svalue` create its own type, i can try doing that

giltho:
> since i opened the PR, i fixed typed.ml to avoid all the ugly type_ etc wrappers, so that flows a lot more nicely; what do you dislike?

I think from your description alone I have a hard time figuring out what it will look like in the end to, for instance, manipulate a fresh Rust enum or something.

This really feel like you're fighting against the type system here, and wondering if it wouldn't be worth designing a "PPX" without all the features we wanted but that basically allows for extending the base definition. At this point it can be its own language and AST, doesn't have to be OCaml. But really what you're solving here is a copy-pasting issue :/

But anyway, ignoring this, maybe if we just spend 20mn on a call and you explain things I'll be very happy with it

N1ark:
Finished upstreaming in #439

PR #425, opened 2026-06-21
Tidy Soteria Rust

N1ark:
Port of some of the improvements and reworks that #418 required

Best reviewed commit by commit, all changes should be small and local

The main changes are
- Changed some code to be type oriented, rather than structural. This means that instead of pattern matching on `rust_val`, we pattern match on the Rust type the `rust_val` is expected to have. This means we are stricter about the well-typedness of our values, and also means we can get rid of some dead code that was hard to spot before :) 
- Made `Value_codec.encode`, `Rtree_block.store` and `Rust_val.Union` keep track of the sizes of the blocks, alongside their offset. This allows us to remove `Value_codec.size_of`, a function I hated!


N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
> 🟡 miri/pass/box-custom-alloc.rs: fail → unsupported

omg this change is actually a bug fix! we used to hackily traverse boxes and replace pointers to remove protectors at function exit, and this broke the allocator if it ever contained a pointer itself, since that pointer would also get replaced! the unsupported is just bc we don't unsize ADTs correctly when the unsized field isn't the last one, but that's ok it's an Obol/Charon issue
```diff
     | TAdt adt when adt_is_box adt ->
         (* a box has only one non ZST, the pointer *)
-        let ptr = as_ptr @@ List.hd @@ flatten v in
+        let ptr, allocator, marker = unwrap_box v in
         let++ ptr, acc = fn init ty ptr in
-        (* recursively look for where the pointer is and replace it *)
-        let rec subst_ptr = function
-          | Tuple vs -> Tuple (List.map subst_ptr vs)
-          | Ptr _ -> Ptr ptr
-          | v -> v
-        in
-        (subst_ptr v, acc)
+        (mk_box (Ptr ptr) allocator marker, acc)
```

soteria-rust/lib/value_codec.ml
```diff
@@ -420,14 +444,25 @@ module Encoder (Sptr : Sptr.S) = struct
...
+            match meta with
+            | Thin -> "unit"
+            | Len _ -> "length"
+            | VTable _ -> "vtable"
+          in
+          let expected =
+            match dst_kind with
+            | NoneKind -> "unit"
+            | LenKind -> "length"
+            | VTableKind -> "vtable"
```
giltho: Surely this should be `pp_dst_kind` implemented somewhere else.
And I'd just implement `pp_meta_kind` as well (which prints the kind without the payload)

giltho (approved):
Huge fan of this, this is *so* pretty, all of this!

PR #426, opened 2026-06-21
Hardcode `Sptr` and `Ptr_tag`

N1ark:
Being parametric on the pointer type was one of the most fun but also most annoying parts of the interpreter: we had to thread the pointer everywhere (for often little benefit).

Let's undo that, as part of #418 

It's just one commit! The only thing it does is:
- Move `Sptr_base` into `sptr.ml`, removing `Sptr.S` since the signature is what's exposed.
- Move `Tree_borrows.Raw.Tag` into `Tree_borrows.Ptr_tag`.

For now I've also gotten rid of all the symbolic machinery around `Ptr_tag`, because we're not using it and once a pointer tag is just a symbolic value it will be much simpler to do. Furthermore, I don't actually think we care about having pointer tags be dependent on the borrows model (for when/if we add Stacked Borrows); in both cases it's just an identifier.

**I strongly recommend reviewing the diff with whitespace changes ignored**, since `Value_codec` lost indentation, as `Value_codec.Encoder` is gone (it was a functor that received a `Sptr`

### DIFF
--- a/soteria-rust/lib/analyses/wpst.ml
+++ b/soteria-rust/lib/analyses/wpst.ml
@@ -98,7 +98,7 @@ let exec_crate (crate : Charon.UllbcAst.crate)
   let@ () = print_outcomes entry_name in
   Fmt.pr "%a %a@." (pp_clr `Teal) "=>" (pp_style `Bold)
     ("Running " ^ entry_name ^ "...");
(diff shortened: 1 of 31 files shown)


giltho (approved):
Extremely pretty

PR #427, opened 2026-06-21
Hide `rust_val` ADT constructor

N1ark:
This is the last PR, before we can make `Rust_val` a typed value (see #418)

This builds on top of #426 to make my life easier, but it can be reviewed on its own (there is only one commit, I dont know how I could have split this); **the only commit this PR adds is 02e8c8e**

The only thing this PR does is define a `rust_val.mli`, that hides the constructors for Rust values, and instead exposes unwrappers and builder to access and modify the values. Don't pay attention to the implementation itself in `rust_val.ml`, as that will change and be replaced. What matters is the interface, and how it affects the rest of the codebase. Note also that right now this introduces extra function calls in places, like `mk_int` and `mk_float`, which will disappear once we merge everything into `Typed`!


soteria-rust/lib/builtins/extern/miri.ml
```diff
@@ -28,21 +28,21 @@ module M (StateM : State.StateM.S) = struct
...
+    | [ ptr ] -> ok (mk_int (Sptr.as_id (fst (as_ptr ptr))))
     | _ -> not_impl "alloc_id: invalid arguments"
 
   let promise_alignement args =
     match args with
-    | [ Ptr (ptr, _); Int align ] ->
-        let align = Typed.cast @@ Typed.cast_i Usize align in
-        let* addr = Sptr.decay ptr in
+    | [ ptr; align ] ->
+        let align = Typed.cast @@ as_base_i Usize align in
```
giltho, line 37: You're forcing align into non-zero, how do you know this holds? Shouldn't you `check_nonzero`?

giltho (approved):
This is going to be so pretty

soteria-rust/lib/builtins/extern/miri.ml
N1ark: indeed ^^' i wont do it here but thats well spotted, ill keep it in mind

PR #341, opened 2026-05-16
OCaml 5.5.0

giltho:
One use of modular explicits! I just really wanted to test it, I know it's not a priority

PR #433, opened 2026-06-24
Distinguish the `test-lib` cache in CI

N1ark:
so the cache for `test-lib` deps was the same as the rest of CI, despite it not actually containing the same things :p this was introduced in #341 which changed the `test-lib` workflow


giltho:
That should be fine though? 
AH WAIT NO, because it contains Soteria in the cache itself 🤦‍♂️

giltho:
I did think about this and thought "no it's fine"

giltho (approved):
👍

PR #434, opened 2026-06-24
Simplify `Foldable`, more monad helpers

N1ark:
As mentioned in #341: add `Monad.fold (module M : Foldable)` (and `iter`) that use modular explicits for the folded type. Unfortunately only having the modular explicit form was actually pretty bad, due to basically all `fold`/`iter` usages causing these sorts of diffs:
```diff
- iter_list pairs ~f:(fun (param, arg) -> S_val.learn_eq param arg)
+ iter
+   (module List)
+   pairs
+   ~f:(fun (param, arg) -> S_val.learn_eq param arg)
```

So i only added the modular explicit form on top, but kept `fold_list`, `fold_iter`, `iter_list`, `iter_iter`.
I removed `foldable.ml` and added the `Foldable` signature to `Sigs`; i instead just tweaked `List` and `Seq` to satisfy `Foldable` directly, avoiding the labelled arguments.

I also spotted a few small performance opportunities in some places where we folded over lists when folding over iterators was sufficient and avoided constructing the list.


N1ark:
!run-benchmarks

giltho (approved):
pretty

PR #437, opened 2026-06-24
Faster CI

N1ark:
Small things I did while not knowing what to do

Disabled `dune-cache` for the `setup-ocaml` jobs, because we install `dune` and keep it in the opam cache anyways, so having it just added ~50s of Dune installation for no reason

Disabled `macos-latest` on PRs, while still running it in the merge queue, on main and for releases. MacOS runners are quite a bit slower and honestly not really needed; it basically never has a different outcome to the ubuntu build.

Skip building JSON docs outside main

### DIFF
--- a/scripts/run_benchmarks.py
+++ b/scripts/run_benchmarks.py
@@ -123,11 +123,10 @@ def bench_rust(entry: dict, kind: str) -> dict:
     soteria_rust = os.environ.get("SOTERIA_RUST", SOTERIA_RUST)
     target = resolve(entry["path"])
     args = entry.get("args", [])
-    base = [soteria_rust, "exec", str(target), *args]
 
     # Compile once (untimed); analysis may exit non-zero on a found bug.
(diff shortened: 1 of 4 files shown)


scripts/run_benchmarks.py
```diff
@@ -123,11 +123,10 @@ def bench_rust(entry: dict, kind: str) -> dict:
     soteria_rust = os.environ.get("SOTERIA_RUST", SOTERIA_RUST)
     target = resolve(entry["path"])
     args = entry.get("args", [])
-    base = [soteria_rust, "exec", str(target), *args]
 
     # Compile once (untimed); analysis may exit non-zero on a found bug.
     log(f"pre-compiling {kind}: {target}")
-    run(base)
```
giltho, line 130: Very important change

giltho (approved):

scripts/run_benchmarks.py
N1ark: taste...

PR #438, opened 2026-06-24
Fix regression in allocator arg parsing

N1ark:
#427 introduced a regression in the `align_of_enum` function for the allocator externs, which caused a big regression in conformance tests. oops!

it also introduced a regression in decoding single-variant enums due to us previously treating those as having an arbitrary layout (rather than an enum layout); i fixed the thing as a whole rather than redoing the workaround


N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

giltho (approved):

PR #440, opened 2026-06-26
Fix CI

N1ark:
the slopper used an expression in an argument description which [crashed the parser](https://github.com/soteria-tools/soteria/actions/runs/28220315320/job/83600561808)


giltho (approved):
The state of linters for github actions is really terrible, how did this pass through any kind of validation hahaha

Issue #176, opened 2025-11-16
Type TreeBlock operations with Nonzero


giltho:
All treeblock lookups should be done with non-zero sizes. It should be caught early

Issue #436, opened 2026-06-24
OX warnings in the stubs


giltho:
We've written quite a lot of stubs now. We've mainly written them assuming UX execution, for instance:

```ocaml
let sysconf args =
  (* https://man7.org/linux/man-pages/man3/sysconf.3.html
   * It is basically ok to always return the i64 `-1` saying "I don't know"
   *)
  ok (mk_int (Typed.BitVec.u64i (-1)))
```

I think it's ok to keep stubs like this for verification, but verification assumptions should be explicitly given to the user using `warn_once`.
We need to comb through all stubs and do the change. 

Possibly, we should have a specific kind of `Warn_once` called `Assumptions`, since they're soundness assumptions idk

PR #442, opened 2026-06-27
Don't allocate ZSTs

### LINKED ISSUE #313: Don't allocate ZSTs
We currently create 0-sized allocations for ZSTs, which is wrong (and probably slightly worsens performance).
Simply "removing" it leads to odd errors so we need to properly look into it and fix this. 

This also is blocking #176

N1ark:
Closes #313 
Enables work on #176 

Skip allocating ZSTs in `Tree_state`; this shouldn't have a massive impact since as of #374 we already return references to ZSTs without allocating, but at least now this is guaranteed!

### DIFF
--- a/soteria-rust/lib/state/tree_state.ml
+++ b/soteria-rust/lib/state/tree_state.ml
@@ -784,17 +784,22 @@ module Make (Borrows : Tree_borrows.T) = struct
     [%l.debug
       "Executing Alloc of size %a (align %a)" Typed.ppa size Typed.ppa align];
     Soteria.Stats.As_ctx.incr StatKeys.allocs;
-    with_heap
-      (let open Heap.SM in
-       let open Heap.SM.Syntax in
-       let*^ block, tag =
-         Freeable_block_with_meta.make ?span ?zeroed ~align ~size ()
-       in
-       let** loc = Heap.alloc ~new_codom:block in
-       let ptr = Typed.Ptr.mk_ptr_t ~loc ~ofs:Usize.(0s) ~tag ~align ~size in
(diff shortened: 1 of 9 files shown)


N1ark:
!run-benchmarks sanity check, but it should be basically the same

soteria-rust/lib/state/tree_state.ml
```diff
@@ -784,17 +784,22 @@ module Make (Borrows : Tree_borrows.T) = struct
...
-         Freeable_block_with_meta.make ?span ?zeroed ~align ~size ()
-       in
-       let** loc = Heap.alloc ~new_codom:block in
-       let ptr = Typed.Ptr.mk_ptr_t ~loc ~ofs:Usize.(0s) ~tag ~align ~size in
-       (* The pointer is necessarily not null *)
-       let+ () = assume [ Typed.(not (Ptr.is_null_loc loc)) ] in
-       ok (Typed.Ptr.mk_ptr_f ptr None))
+    if%sat size ==@ Usize.(0s) then
+      (* UX: we under-approximate and assume the address is align, though it can
+         in fact be any well-aligned address. *)
```
giltho, line 789: Hmmm, #436 only mentions stubs. Can we agree on some comment so we can search-replace easily?

giltho (approved):
Minor about commentcs

soteria-rust/lib/state/tree_state.ml
N1ark: `UX: some comment` it is

N1ark: added it in a few places

PR #445, opened 2026-06-30
Upstream fixes and minor features from the CN work

giltho:
- Removes `soteria-c/lib/state.mli`. There's no point in hiding the interface (and I added a module type-check somewhere else).
- Adds `Symex.log_solver_state ~level ()`
- Adds `Sigs.Eq`
- Fix `Dirty_vars` not being pulled in. If I have `Dirty_vars { x , y }` and I know that `x` is relevant, then I should also add `y` to my set of relevant variables (otherwise transitive equalities are lost...)
- Adds `take_nth` to `Soteria_std.List`
- Fix memoisation tables to keep the values, not just their tags 😬
- Add an implementation of the Lazy map. **Maybe look at this one in details**, I've made implementation choices :p
- Fixed union find "default" value being mutable yet copied in every fresh solver
- Fixes the S_map being very wrong when doing the member check 🤦‍♂️, and adds a few tests for the lazy map behaviour
- Fix map distinct and change the default of mk_commut_nop to be no-idempotent by default (the least risky choice)

### DIFF
--- a/soteria/lib/bv_values/analyses.ml
+++ b/soteria/lib/bv_values/analyses.ml
@@ -252,7 +252,7 @@ module Make (Typed : Typed_intf.Solver_value) = struct
       type t = st
 
       let default () = Var.Map.empty
-      let copy t = t
+      let copy = Fun.id
     end)
 
     let get n v st =
--- a/soteria/lib/bv_values/analyses.ml
+++ b/soteria/lib/bv_values/analyses.ml
@@ -251,7 +251,8 @@ module Make (Typed : Typed_intf.Solver_value) = struct
     include Reversible.Make_mutable (struct
       type t = st
 
-      let default = Var.Map.empty
+      let default () = Var.Map.empty
+      let copy t = t
     end)
 
     let get n v st =
@@ -558,13 +559,9 @@ module Make (Typed : Typed_intf.Solver_value) = struct
     include Reversible.Make_mutable (struct
       type t = Svalue.t UnionFind.store * Svalue.t UnionFind.rref VMap.t
 
-      let default = (UnionFind.new_store (), VMap.empty)
+      let default () = (UnionFind.new_store (), VMap.empty)
+      let copy (uf, refs) = (UnionFind.copy uf, refs)
     end)
-
-    (* override to account for UnionFind *)
-    let save d =
-      let uf, st = Dynarray.get_last d in
-      Dynarray.add_last d (UnionFind.copy uf, st)
 
     let rec cost (v : Svalue.t) : int =
       match v.node.kind with
--- a/soteria/lib/bv_values/svalue.ml
+++ b/soteria/lib/bv_values/svalue.ml
@@ -514,6 +514,13 @@ module Make (V : Value_ext) () = struct
       | Extension e -> combine (combine h 12) (V.hash e)
   end)
 
+  module Hashtbl = Hashtbl.Make (struct
+    type nonrec t = t
+
+    let equal = equal
+    let hash = hash
+  end)
+
   let ( <| ) kind ty : t = Hcons.hashcons { kind; ty }
   let mk_var v ty = Var v <| ty
 
--- a/soteria/lib/bv_values/svalue.ml
+++ b/soteria/lib/bv_values/svalue.ml
@@ -532,7 +532,7 @@ module Make (V : Value_ext) () = struct
   (** We put commutative n-ary operators in some sort of normal form where
       elements are sorted by ud, to increase cache hits. If [idem] is true, will
       also remove duplicates. *)
-  let mk_commut_nop ?(idem = true) op vs =
+  let mk_commut_nop ~idem op vs =
     let sort = if idem then List.sort_uniq else List.sort in
     let vs = sort (fun l r -> Int.compare l.tag r.tag) vs in
     Nop (op, vs)
@@ -1099,6 +1099,7 @@ module Make (V : Value_ext) () = struct
           split_ands s2 f
       | _ -> f sv
 
+    (* If l is not none, it is the list represented by the sequence S *)
     let distinct_raw s ?l () =
       (* [Distinct l] when l is empty or of size 1 is always true *)
       match Seq.compare_length_with s 2 with
@@ -1117,8 +1118,9 @@ module Make (V : Value_ext) () = struct
           match (res, l) with
           | Some true, _ -> v_true
           | Some false, _ -> v_false
-          | None, Some l -> mk_commut_nop Distinct l <| TBool
-          | None, None -> mk_commut_nop Distinct (List.of_seq s) <| TBool)
+          | None, Some l -> mk_commut_nop ~idem:false Distinct l <| TBool
+          | None, None ->
+              mk_commut_nop ~idem:false Distinct (List.of_seq s) <| TBool)
 
     let distinct_seq s = distinct_raw s ()
     let distinct l = distinct_raw (List.to_seq l) ~l ()
--- a/soteria/lib/data/s_map.ml
+++ b/soteria/lib/data/s_map.ml
@@ -111,7 +111,9 @@ struct
     match M.find_opt key st with
     | Some v -> Symex.return (key, Some v)
     | None ->
-        let not_in_map = M.to_seq st |> Seq.map fst |> Key.distinct_seq in
+        let not_in_map =
+          M.to_seq st |> Seq.map fst |> Seq.cons key |> Key.distinct_seq
+        in
         B.if_not_in_map not_in_map
           ~then_:(fun () -> Symex.return (key, None))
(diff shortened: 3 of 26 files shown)


soteria/lib/bv_values/analyses.ml
```diff
@@ -251,7 +251,8 @@ module Make (Typed : Typed_intf.Solver_value) = struct
     include Reversible.Make_mutable (struct
       type t = st
 
-      let default = Var.Map.empty
+      let default () = Var.Map.empty
+      let copy t = t
```
N1ark: i wonder if for these sorts of functions (we have quite a few) there is any perf improvement in instead doing `let copy = Fun.id` or maybe `let copy = "%identity". like idk if maybe of these could help the compiler better than us just defining it as `x = x`
guess i should try it out :P

soteria/lib/bv_values/svalue.ml
```diff
@@ -525,7 +532,7 @@ module Make (V : Value_ext) () = struct
   (** We put commutative n-ary operators in some sort of normal form where
       elements are sorted by ud, to increase cache hits. If [idem] is true, will
       also remove duplicates. *)
-  let mk_commut_nop ?(idem = true) op vs =
+  let mk_commut_nop ?(idem = false) op vs =
```
N1ark: can we just
```suggestion
  let mk_commut_nop ~idem op vs =
```

soteria/lib/bv_values/svalue.ml
```diff
@@ -1097,21 +1104,29 @@ module Make (V : Value_ext) () = struct
...
-          match (res, l) with
-          | Some true, _ -> v_true
-          | Some false, _ -> v_false
-          | None, Some l -> mk_commut_nop Distinct l <| TBool
-          | None, None -> mk_commut_nop Distinct (List.of_seq s) <| TBool)
+          match l with
+          | Some l ->
+              let length = List.length l in
+              let sorted = List.sort_uniq compare l in
+              if List.length sorted <> length then v_false
```
N1ark: `List.compare_length_with` or whatever it's called

soteria/lib/bv_values/svalue.ml
```diff
@@ -1097,21 +1104,29 @@ module Make (V : Value_ext) () = struct
...
+                Nop (Distinct, sorted) <| TBool
+          | None ->
+              let a = Array.of_seq s in
+              Array.fast_sort (fun v1 v2 -> v1.tag - v2.tag) a;
+              let exception FoundDuplicate in
+              let duplicates =
+                try
+                  Array.iteri
+                    (fun i v ->
+                      if i > 0 && equal a.(i - 1) v then raise FoundDuplicate)
```
```suggestion
                      if i > 0 && equal a.(i - 1) v then raise_notrace FoundDuplicate)
```

soteria/lib/bv_values/svalue.ml
```diff
@@ -1097,21 +1104,29 @@ module Make (V : Value_ext) () = struct
...
+                    a;
+                  false
+                with FoundDuplicate -> true
+              in
+              if duplicates then v_false
+              else mk_commut_nop Distinct (List.of_seq s) <| TBool)
```
N1ark: the formatting is gonna be wrong but you get it
```suggestion
                  mk_commut_nop Distinct (List.of_seq s) <| TBool)
                with FoundDuplicate -> v_false
```

soteria/lib/data/s_map.ml
```diff
@@ -102,9 +111,61 @@ struct
...
+      (Symex)
+      (struct
+        open Symex
+
+        (** In the direct access map, we check if it's possible that the key is
+            outside the map, and if it is, we only take that branch in UX.
+            However, we keep the information that the value is not in the map,
+            we have entirely dropped the other branch. *)
+        let if_not_in_map guard ~then_ ~else_ =
+          branch_on_take_one guard ~then_ ~else_
```
N1ark, line 138: does `if_not_in_map = branch_on_take_one` not work because of weak vars?

soteria/lib/data/s_map.mli
```diff
@@ -36,6 +37,14 @@ module Direct_access_patricia_tree
     (Symex : Symex.Base)
     (Key : Key(Symex).S_patricia_tree) : S(Symex)(Key).S
 
+(** Makes a symbolic map similar to {!Direct_access}, but even in OX, if it is
+    possible that the key is outside the map, only this path will be taken. *)
+module Lazy (Symex : Symex.Base) (Key : Key(Symex).S) : S(Symex)(Key).S
+
+module Lazy_patricia_tree
```
N1ark, line 47:
```suggestion
(** Same as {!Lazy}, but backed by a
    {{:https://ocaml.org/p/patricia-tree/latest/doc/index.html}Patricia Tree},
    which may offer performance benefits. *)
module Lazy_patricia_tree
```

soteria/lib/data/s_map_intf.ml
```diff
@@ -48,6 +48,10 @@ struct
   module type S = sig
     type 'a t
 
+    (** Indicates whether the map has a lazy behaviour. If true, it means that:
+        - finding a hit does not mean it is the only possible hit, and
+        - not finding a hit does not mean there is no match. *)
+
     val empty : 'a t
```
N1ark: what

soteria/lib/soteria_std/list.ml
```diff
@@ -170,6 +170,19 @@ let rec find_with_rest f l =
...
+  let rec aux n l =
+    match (n, l) with
+    | _, [] -> None
+    | 0, x :: xs -> Some (x, xs)
+    | n, x :: xs -> (
+        match take_nth (n - 1) xs with
+        | None -> None
+        | Some (found, rest) -> Some (found, x :: rest))
+  in
+  aux n l
```
N1ark: do we really want the `option`? to me this sort of thing should just raise an `Invalid_argument` if out of bounds...? not sure where you use it

soteria/lib/symex/symex.ml
```diff
@@ -171,6 +171,7 @@ module type Core = sig
   (** {2 Fuel} *)
 
   val consume_fuel_steps : int -> unit t
+  val log_solver_state : level:Logs.Level.t -> unit -> unit
```
N1ark, line 176: took me a second to understand ngl
```suggestion

  (** Logs the current solver state, at the given log level *)
  val log_solver_state : level:Logs.Level.t -> unit -> unit
```

N1ark (changes requested):
few nits on `distinct` + doc

soteria/lib/data/s_map.ml
giltho: Yep
Also, optional arguments need to be explicitly erased by passing at least the guard

soteria/lib/data/s_map_intf.ml
giltho: Woops, I ended up removing that and didn't kill the doc

soteria/lib/bv_values/svalue.ml
giltho: I'm actually reverting the new definition of distinct. It doesn't reduce concrete stuff. We should optimise for concrete stuff to be removed, and if we're not sure then we stop which is what the old did

soteria/lib/soteria_std/list.ml
giltho: Hmm, sure

giltho: The issue is that `find_with_rest` above returns an option

N1ark: well because `find` is about "maybe this doesn't exist", because its a predicate, it's much harder to have a guarantee

if you have an index usually you know the index exists

giltho: renamed it take_nth_exn

soteria/lib/bv_values/analyses.ml
giltho: I wanna say "surely the compiler knows to do that" but 🤷‍♂️ lmao

soteria/lib/bv_values/svalue.ml
```diff
@@ -1092,26 +1099,25 @@ module Make (V : Value_ext) () = struct
           split_ands s2 f
       | _ -> f sv
 
+    (* If l is not none, it is the list represented by the sequence S *)
     let distinct_raw s ?l () =
       (* [Distinct l] when l is empty or of size 1 is always true *)
-      match Seq.compare_length_with s 2 with
-      | -1 -> v_true
-      | _ -> (
```
N1ark, line 1099: hmm but now we lost the reduction that `distinct(x)` is `true` ... : (

N1ark (approved):

soteria/lib/bv_values/svalue.ml
giltho: aaaaaaa mabad

PR #446, opened 2026-07-02
Fix hack in decoding values

N1ark:
- Removed the hack in `with_ptr` that would catch misses and convert them to dangling pointer errors if the pointer has no provenance; we now instead properly check whether the pointer is null (null deref) or has no provenance (dangling pointer), meaning `Rtree_block` shouldn't miss anymore
- Move the decoder handling logic inside `Rtree_block`, to avoid having to go through `Heap.with_ptr` (PMap access) + `Block.with_block` on any memory access. This is a nice 10% improvement on a test I had that used a sysroot!
- Made sure no-op heap accesses (e.g. reading a ZST, getting the discriminant of an enum with a known discriminant) never go through the heap

the perf results are a bit mid but i know for sure that for any heap access that requires >3 tree block operations, this is better

### DIFF
--- a/soteria-rust/lib/layout.ml
+++ b/soteria-rust/lib/layout.ml
@@ -442,6 +442,15 @@ let[@inline] is_zst ty =
   let++ layout = layout_of ty in
   layout.size ==@ Usize.(0s)
 
+(** If the given type is an enum with a single variant, return the variant's ID.
+    Otherwise, return None. Fails if the type is not an enum. *)
+let enum_single_variant (ty : Types.ty) =
+  let++ layout = layout_of ty in
+  match layout.fields with
+  | Enum (Known v, _) -> Some v
+  | Enum _ -> None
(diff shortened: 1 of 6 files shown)


N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
hmmm odd results; will investigate..........

N1ark:
ok so this is because even if we defer the null ptr / dangling check, accessing the pmap in `with_ptr` will check if the block has been freed. this is a lot nicer aesthetically so i'd really want to see if i can salvage this approach regardless :/

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

giltho (approved):

PR #449, opened 2026-07-03
Disable recursive validity by default

N1ark:
Given the decision taken here (https://github.com/rust-lang/unsafe-code-guidelines/issues/414), contents of memory will not matter for reference validity, i.e. our `--recursive-validity` check is not necessary. I don't want to remove it because it's there and someone might want it, but let's disable it by default :) 

It's this check that caused a 9086.2% regression on the `array_init.rs` benchmark on #391 ; the full 1024 long array was being decoded at every reference projection to check for validity. I assume some inlining that happens when there is no sysroot prevented a reference to the full array to be passed around, meaning that the validity check was seldom performed on the full thing

This should also be a significant perf improvement!


N1ark:
!run-benchmarks

soteria-rust/lib/config.ml
```diff
@@ -159,10 +159,13 @@ type t = {
...
       [@docs Sections.analysis]
-      [@default Warn]
+      [@default Allow]
       [@names [ "recursive-validity" ]]
       (** Whether to check the validity of the addressed memory when obtaining a
-          reference to it. We only go one level deep. *)
+          reference to it. We only go one level deep. This does not cause UB,
+          but may cause UB down the line. This may also have a significant
+          impact on performance. For more information, see
+          https://github.com/rust-lang/unsafe-code-guidelines/issues/414 *)
```
giltho: Pedantic here, but "This does no cause UB... This may have significant impact on performance..."
They're not the same this? Or I'm confused?

I'd write:
```
Referencing invalid values is not considered UB, but is bad practice as it may cause UB down the line. Enabling recursive validity checks may have significant impact on performance.
```

soteria-rust/test/cram/cli.t/run.t
```diff
@@ -220,9 +220,12 @@
              The provenance model to use for pointers. If not provided, the
              default is permissive.
   
-         --recursive-validity=ENUM (absent=warn)
+         --recursive-validity=ENUM (absent=allow)
```
giltho: nitpick that isn't new but `--recursive-validity=allow` seems to imply you're allowing recursive validity.
Can we rename this flag `--references-to-invalid-memory`?

giltho (approved):

soteria-rust/test/cram/cli.t/run.t
N1ark: sure :P

N1ark: actually idk, Miri uses `-Zmiri-recursive-validation` i think it makes sense to keep it similar, if anything id say we change it to `--recursive-validation`

giltho: But then it's a boolean, not allow/warn/disallow?
I'd say pick one, either bool or --reference-to-invalid-memory

PR #450, opened 2026-07-06
Use nonzero values in TreeBlock 

### LINKED ISSUE #176: Type TreeBlock operations with Nonzero

N1ark:
Closes #176

### DIFF
--- a/soteria-c/lib/ctree_block.ml
+++ b/soteria-c/lib/ctree_block.ml
@@ -213,12 +213,12 @@ include Tree_block (MemVal)
 module Range = struct
   include Range
 
+  let of_low_and_size low (size : Typed.([< T.nonzero ] t)) =
+    of_low_and_size low (size :> Typed.(T.sint t))
+
   let of_low_and_type low ty =
     let+ size = Layout.size_of_s ty in
     of_low_and_size low size
-
-  let of_low_and_size low (size : Typed.([< T.nonzero ] t)) =
-    of_low_and_size low (size :> Typed.(T.sint t))
 
   let size r = Typed.BitVec.cast_nonzero @@ size r
 end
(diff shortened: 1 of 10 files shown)


soteria-c/lib/ctree_block.ml
```diff
@@ -274,14 +284,15 @@ let store (low : [< T.sint ] Typed.t) (ty : Ctype.ctype)
         match node.node with
         | NotOwned Totally ->
             let* len = Layout.size_of_s ty in
```
giltho, line 286: Let's make "size_of" return a non-zero? It's always the case in C

soteria-rust/lib/builtins/intrinsics/impl.ml
```diff
@@ -1043,7 +1043,7 @@ module M (StateM : State.StateM.S) : Intf.M(StateM).Impl = struct
      *   handle the symbolic size case yet.
      *)
     if%sat size ==@ zero then ok ()
```
giltho, line 1045: I like to know where my assumptions come from and make them "type safe"
I'd prefer if you didn't cast inside the expression

I prefer
```
if%sat size ==@ zero then ok ()
else
   (* we just checked *)
   let size = cast_nonzero size in
   ...
```

soteria-c/lib/stubs.ml
```diff
@@ -211,7 +211,9 @@ module M (State : State_intf.S) = struct
       | _ -> not_impl "memset with non-thre arguments"
     in
     let char = BV.fit_to ~signed:false 8 char_int in
-    if%sure char ==@ U8.(0s) then
+    if%sat count ==@ Usize.(0s) then Result.ok Agv.void
+    else if%sure char ==@ U8.(0s) then
+      let count = Typed.BitVec.cast_nonzero count in
```
giltho: Same as above

giltho (approved):

soteria-rust/lib/builtins/intrinsics/impl.ml
N1ark: yeah but this adds a whole level on indentation to all the code below this, and it's not needed anywhere else; in general i would agree but for here its just substantially uglier </3

PR #452, opened 2026-07-07
Minor store reorganisation

N1ark:
Minor change to Soteria Rust's store, to get rid of `bind_value` and make `try_load` more sensible and more similar to `try_store`. I always disliked how different `try_load` was from `try_store`, and how `bind_value` worked which was kind of weird, figured out a way of simplifying it

This also makes `Store.Place.read_val` reusable for the heap abstraction, I think :)


N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

giltho (approved):
Absolutely

PR #455, opened 2026-07-14
Minor Tree Borrows improvements 

N1ark:
done while working on the blog post

main change is also adding a strongmap/strongset to make it easier to swap around, and tweaking the way we Gc


N1ark:
!run-benchmarks

giltho (approved):
yeah

PR #456, opened 2026-07-22
Heap abstraction

### LINKED ISSUE #163: Best effort heap abstraction
Similarly to #162, even when we store values in the heap we may not always need to use treeblocks or to split them into smaller chunks. We may allow the encoding to store values into non-atomic parts (e.g. a struct as a whole), and only decompose them if needed. 
Similarly, loads/stores could read/modify the fields in the struct directly rather than splitting the tree block, maybe.

N1ark:
they thought it could never be done...

add a `Whole of Typed.(T.any t) * Types.ty` variant to `Rtree_block`, to represent non-scalar leaf values. By default, storing a value always stores it whole, and further reads may split the value as much as needed to get to what they need. When splitting, we only split one depth at a time, rather than decaying the entire value!

Closes #163


N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -47,7 +47,20 @@ module Make (Borrows : Tree_borrows.M(DecayMap.SM).S) = struct
...
-    type leaf = Init of Typed.(T.any t) | Zeros | Uninit | Any | Unowned
+
+    type leaf =
+      | Init of Typed.(T.any t)
+          (** A leaf value: either a [TBitVector], [TFloat] or [TFullPtr] with
+              no metadata. *)
+      | Whole of Typed.(T.any t) * Types.ty
+          (** A whole aggregate value, stored unencoded along with its
+              (normalised) type; it is only (shallowly) decomposed when an
+              access requires it. *)
```
giltho: Why do you need these two to be different variants?

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -78,47 +93,102 @@ module Make (Borrows : Tree_borrows.M(DecayMap.SM).S) = struct
...
       | _, _ -> Lazy
 
-    let rec split_rval (v : Typed.([< T.any ] t)) at =
+    (** Converts a scalar to its bitvector representation. *)
+    let scalar_to_bv (v : Typed.([< T.any ] t)) =
       match%ty v with
-      | TExtension TFullPtr -> (
+      | TBitVector _ -> return v
+      | TExtension TFullPtr ->
           let ptr, meta = Typed.Ptr.split v in
```
giltho, line 100: I know it's orthogonal to this PR but I find the name `Typed.Ptr.split` very ambiguous, I'd expect it to split loc from offset. I stopped for a second when reading this.
How about `thin_and_meta`?

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -78,47 +93,102 @@ module Make (Borrows : Tree_borrows.M(DecayMap.SM).S) = struct
...
+      (mask_l, mask_r)
+
+    (** Builds a right-leaning split tree out of subtrees paired with the offset
+        of their (exclusive) upper end; [base] is the offset at which the
+        subtree starts. *)
+    let rec mk_split_tree base = function
+      | [] -> L.failwith "mk_split_tree: no parts"
+      | [ (t, _) ] -> t
+      | (t, high) :: rest ->
+          TB.Split_tree.Node (t, high -!!@ base, mk_split_tree high rest)
```
giltho, line 135: Such a weird function, why do you need this?

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -78,47 +93,102 @@ module Make (Borrows : Tree_borrows.M(DecayMap.SM).S) = struct
...
+              let+ rest = insert p rest in
+              p' :: rest
+      in
+      let rec sort = function
+        | [] -> return []
+        | p :: rest ->
+            let* rest = sort rest in
+            insert p rest
+      in
+      sort parts
```
giltho: Data.List.Semi_concrete.sort?

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -78,47 +93,102 @@ module Make (Borrows : Tree_borrows.M(DecayMap.SM).S) = struct
...
-          let mask_l = BV.extract 0 ((at * 8) - 1) v in
-          let mask_r = BV.extract (at * 8) ((size * 8) - 1) v in
-          (mask_l, mask_r)
-      | _ -> not_impl "Split unsupported: %a at %a" Typed.ppa v Typed.ppa at
+          assert (Option.is_none meta);
+          Sptr.decay ptr
+      | TFloat _ -> Value_codec.float_to_bv_bits v
+      | _ -> not_impl "Unexpected value in lazy decoding: %a" Typed.ppa v
+
+    let split_rval (v : Typed.([< T.any ] t)) at =
```
giltho: This function is at least badly named. It should be `split_scalar`.
But really I'd rather you have a single `split_rval` function where `rval_to_bv` returns an option and depending on being able to get a bv immediately, you split if you can, or you explode and try again if you can't.

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -78,47 +93,102 @@ module Make (Borrows : Tree_borrows.M(DecayMap.SM).S) = struct
...
+      let size = Typed.size_of_int v / 8 in
+      let+ at =
+        match BV.to_z at with
+        | Some at -> return (Z.to_int at)
+        | _ -> (
+            (* HACK: we need to branch on the concrete size, because the actual
+               bitvector sort of the value must have a concrete size.
+
+               As per the contract of [split], we know [at ∈ [1, size)] *)
+            let options = List.init (size - 1) (( + ) 1) in
```
giltho, line 118: Slowly sliding down the point-free slope

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -215,15 +314,25 @@ module Make (Borrows : Tree_borrows.M(DecayMap.SM).S) = struct
             let+ () = learn_eq e v in
             Unowned
         | SInit _, Zeros -> lift @@ not_impl "Assume rust_val.syn == 0s"
+        | SInit _, Whole _ | SWhole _, Init _ ->
+            lift @@ not_impl "consume whole/init mismatch"
```
giltho: Convert the init to a whole and consume?

soteria/lib/soteria_std/list.ml
```diff
@@ -213,6 +213,13 @@ let split_around l i =
   in
   aux [] l i
 
+(** [split_hd l] returns the head and tail of the list [l].
+
+    @raise Invalid_argument if [l] is empty. *)
+let[@inline] split_hd = function
```
giltho:
```suggestion
let[@inline] take_first = function
```

giltho (changes requested):
My main concern is the separation between Whole and Init. I don't like it, and I don't understand why it's necessary. I think it will make other things much more complex for bi-abduction later.

soteria-rust/lib/state/rtree_block.ml
N1ark: split is getting removed in #462 so don't worry hehe

soteria-rust/lib/state/rtree_block.ml
N1ark: i mean its just a small helper? `MemVal.split` must return `Split_tree`s for either sides, and once we're done splitting the `Whole` into a list of values we only have a list, so we need to recontruct a tree. That's just part of the interface

soteria-rust/lib/state/rtree_block.ml
N1ark: hehe .. tbh here it could just be `Int.succ` that's my bad

soteria-rust/lib/state/rtree_block.ml
N1ark: yeah whatever todos there are in cons/prod it's just that i would rather not bother just now and do it later, since its not used anywhere in soteria rust since we dont do compositionality here

soteria-rust/lib/state/rtree_block.ml
N1ark: discussed on discord -- renamed to `Scalar | Aggregate`

soteria-rust/lib/state/rtree_block.ml
N1ark: renamed to `split_scalar`

giltho (approved):

PR #457, opened 2026-07-23
Use `ppx_matches`

N1ark:
i love small codeeee


soteria-rust/lib/state/tree_state.ml
```diff
@@ -302,9 +301,8 @@ module Make (Borrows : Tree_borrows.T) = struct
 
   let pp_pretty ~ignore_freed ft { heap; _ } =
     let (ignore : 'a * Freeable_block_with_meta.t -> bool) =
-      if ignore_freed then function
-        | _, { node = Freed; _ } -> true | _ -> false
-      else fun _ -> false
+      if ignore_freed then [%matches? _, { node = Freed; _ }]
+      else Fun.const false
```
giltho: Hating `Fun.const`. Same amount of characters, obfuscates the meaning.
Please revert to `fun  _ -> false`

giltho (approved):
Please yeet `Fun.const` and that's about it.

PR #458, opened 2026-07-24
Encode enum variants concretely

N1ark:
In the process of encoding Rust values into SMT, I realised that our current system of having symbolic discriminants is incompatible with SMT-encoding. Given an enum value, the encoder must know what variant it is, without performing any symex. This is of course not possible with a symbolic discriminant! This was always kind of hacky anyways: an enum also carries its list of values, and that list is concrete, so this only really worked for some edge cases.

This simplifies a bit of code; the only cost is that the `Ord` binop now branches if multiple results are possible. This will be fixed when we have proper symbolic enums though, so it's a temporary regression.

```diff
-  | Enum of 'ghost sv * 'ghost sv list  (** discriminant * values *)
+  | Enum of Types.variant_id * ''ghost sv list (** variant * values *)
```

As a result, the natural operation to have isn't `discriminant_of` anymore but `is_variant`! `discriminant_of` is now just expressed in terms of `is_variant` and `ite`s -- this will make life easier when we do SMT encoding.

ill probably write somewhere a tutorial on what im learning from this process and some best practices when making extensions

### DIFF
--- a/soteria-rust/lib/builtins/core.ml
+++ b/soteria-rust/lib/builtins/core.ml
@@ -14,12 +14,28 @@ module M (StateM : State.StateM.S) = struct
   open Syntax
 
   let cmp ~signed l r =
+    let open Typed.Adt.Checked in
     let ordering = Crate.get_adt_lang_item_ref RustcLangItemOrderingEnum in
     let ( < ) = if signed then ( <$@ ) else ( <@ ) in
-    let discr =
-      Typed.ite (l < r) U8.(-1s) (Typed.ite (l ==@ r) U8.(0s) U8.(1s))
-    in
-    Typed.Adt.mk_enum ordering discr []
+    let lt = l < r in
+    let eq = l ==@ r in
+    if%sat lt then ok @@ mk_enum ordering "Less" []
+    else if%sat eq then ok @@ mk_enum ordering "Equal" []
+    else ok @@ mk_enum ordering "Greater" []
+  (* TODO: uncomment this when we support symbolic enums, and remove the
+   * three lines above
+   *
+   * match (Typed.to_bool lt, Typed.to_bool eq) with
+   * | Some true, _ -> ok @@ mk_enum ordering "Less" []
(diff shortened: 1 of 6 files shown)


soteria-rust/lib/builtins/core.ml
```diff
@@ -14,12 +14,25 @@ module M (StateM : State.StateM.S) = struct
...
+    | _ ->
+        (* TODO: uncomment this when we support symbolic enums
+         * let*^ ord = Rustsymex.nondet (Typed.t_enum ordering) in
+         * let discr = Typed.Adt.discriminant_of ord in
+         * let exp_discr = Typed.ite lt U8.(-1s) (Typed.ite eq U8.(0s) U8.(1s)) in
+         * let+ () = assume [ discr ==@ exp_discr ] in
+         * (ord : T.enum Typed.t :> [> T.enum ] Typed.t) *)
+        if%sat lt then ok @@ mk_enum ordering "Less" []
+        else if%sat eq then ok @@ mk_enum ordering "Equal" []
+        else ok @@ mk_enum ordering "Greater" []
```
giltho: This feels like some abstraction should catch that repetition.
The code without the comments is equivalent to just the if%sats. Adding that code in one of the if%sats should somehow magically work, I need to think about it.

giltho (approved):

soteria-rust/lib/builtins/core.ml
N1ark: yes, you are correct that as the code stands those three bottom lines are entirely sufficient, and the top part is useless duplication. happy to clean it up here. the reason they're there is that when i will remove the `if%sat`s and uncomment the code, i want to not create a `nondet` value in the case where it can be solved concretely. 

i could also use this to avoid the pattern match, (once i can uncomment that code) if youd prefer?
```ocaml
if%sure lt then <concrete> 
else if%sure eq then <concrete> 
else if%sure neither then <concrete> 
else <nondet + assume constraint>
```

Issue #421, opened 2026-06-20
Make Rust values solver-compatible


N1ark:
As done in #418, I just need to clean that up and split it.

- [x] #422
- [x] #425 
- [x] #426 
- [x] #427 
- [x] #439 

After that, we need some additional work:
- [x] #460 
- [ ] #464 !

PR #460, opened 2026-07-26
Add solver declarations

N1ark:
Add `Soteria.Solvers.Decls`, which allows value encodings to add declarations to the current scope, through an algebraic effect. Declarations are keyed and lazily computed: if two values in the same scope need the same declaration, its SMT encoding will only be created and sent to the solver once.

This is all done within `Solvers`, separately from `Soteria.Symex`, since it is solver-dependent.

Thanks to this, we can now also provide an encoding for `Bv_values`'s  pointers! this used to not be possible due to the need to make declarations that were dependent on the pointer size. This also will enable encoding Rust enums into SMT-LIB :) one step closer to #421 

im quite happy with the design of this but if you have better ideas i welcome them; i could have make the embedding a bit deeper i suppose rather than raising straight `sexp`s but i think this is simpler to use. i also contemplated adding a `Decls.declare_datatype` convenience that used the datatype's name as a key, but that would require evaluating all parts of the datatype eagerly, which seems like an anti-pattern when we're providing a lazy iterator

### DIFF
--- a/soteria/lib/solvers/decls.ml
+++ b/soteria/lib/solvers/decls.ml
@@ -1,0 +1,37 @@
+(** Auxiliary SMT-LIB declarations required by value encodings.
+
+    A {{!Value.S}solver value} (or in some cases, a
+    {{!Soteria.Bv_values.Svalue.Value_ext}value extension} may need global
+    SMT-LIB declarations to support its encoding, e.g. declaring an algebraic
+    datatype for a recursive value type. Encoders cannot communicate with the
+    solver directly, so we instead provide a function, {!declare}, to allow
+    doing it.
+
+    Solver implementations (like {!Z3}) handle the raised effect, {!Declare}, by
+    de-duplicating declarations (by {!field:key}) and sending them to the
+    underlying solver process before the command that is about to be run.
+    {b There is no guarantee the declaration will still be in scope for whatever
+       command follows:} encoders must {e always} re-declare definitions they
+    may need. *)
+
+type t = {
+  key : string;
+      (** Canonical identifier for the declaration. Two {!declare}s with the
+          same [key] are assumed to produce identical [commands], and are
+          deduplicated. *)
+  commands : Smt.sexp Iter.t;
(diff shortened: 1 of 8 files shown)


soteria/lib/solvers/decls.ml
```diff
@@ -0,0 +1,37 @@
...
+
+type t = {
+  key : string;
+      (** Canonical identifier for the declaration. Two {!declare}s with the
+          same [key] are assumed to produce identical [commands], and are
+          deduplicated. *)
+  commands : Smt.sexp Iter.t;
+      (** Lazily produces the SMT-LIB commands declaring the object(s); only
+          invoked if [key] is not already declared, so that the declaration is
+          never built otherwise. *)
```
giltho, line 25: Wouldn't a `string lazy_t` be better?
Like the already fully-encoded and ready to be sent through the pipe?

giltho (approved):
Just that thing, what's your opinion?

soteria/lib/solvers/decls.ml
N1ark: yeah sure ! we don't really use lazy anywhere so i didn't think about it

N1ark: nevermind; we cant use Lazy; the declaration may itself recursively require further declarations, and if we use lazy then those dependencies aren't going to be redeclared. an option is exposing a writer monad that keeps track of declarations, so that the computation is actually done once and remembers all dependencies, but that seems super heavy-weight

giltho: Another option is to get it in a cache table? You have a map key -> declaration, since declarations are unique per key. But also, if lazy doesn't work, let's just merge as-is.

N1ark: yeah but you'd need map key -> declaration * map key set for the dependencies, and at that point you're just building the declaration DAG (which could be interesting! just seems like a lot of work for something that just Works as is); i doubt reencoding the expression into a sexp is what's costing us anyways, since that code is reached when we use z3

PR #462, opened 2026-07-27
Force typedness of pointer metadata

N1ark:
While working on the encoding of pointers, I realised the current `meta_of` function is super painful to implement symbolically; it basically requires having symbolic options, which id want to avoid.

Luckily for us, our engine is well-typed, so we actually never really need `meta_of`! Instead, we rely on `dst_kind`, the fat pointer kind of types, and directly retrieve the right metadata type. When we used `meta_of` just to retrieve the metadata and put it back into the modified pointer, we now provide `Ptr.with_ptr : full_ptr -> thin_ptr -> full_ptr`.

For similar reasons, we also need to stop encoding pointers as having a `Svalue.t option` metadata, as if the metadata is taken from a symbolic pointer we don't have any `option` to go with. Instead, we add a `PtrMeta` value type and kind, which contains the (possibly unit) metadata. This type is **never leaked to the engine**: it only lives inside `ext.ml`/`typed.ml`, and does not reach the rest of the codebase. It is only to provide a uniform type for both concrete and symbolic ptr meta.

To go with the above change, tweaked `mk_ptr_f : ptr -> meta option -> fptr` to instead be `ptr -> meta -> fptr`, where meta is `[< sint | ptr ]`. To create a fat pointer with unit metadata, we then provide `of_ptr_t : ptr -> fptr`. This makes calling code a bit simpler as we avoid an unnecessary option :)  

The change is a bit churny and requires normalising types (since `dst_kind` relies on it) but it seems to work pretty well. Don't worry too much about the implementation in `typed.ml`, as it will be replaced by a symbolic-value implementation very soon


N1ark:
!run-benchmarks this supposedly is not observable

N1ark:
!run-benchmarks love when ci decides not to run for reasons

N1ark:
!run-benchmarks how about now

soteria-rust/lib/builtins/fixme/impl.ml
```diff
@@ -29,9 +29,9 @@ module M (StateM : State.StateM.S) : Intf.M(StateM).S = struct
         Usize.(2s)
         vtable
     in
-    let align_ptr = Typed.Ptr.mk_ptr_f align_ptr None in
+    let align_ptr = Typed.Ptr.of_ptr_t align_ptr in
```
giltho, line 32: Minor: but I'm not super happy with the various function and module names.
Maybe `Typed.FatPtr.of_thin`? Or just `Ptr.of_thin`?

giltho: AH reading, `of_ptr_t`, the `t` is not `type` but `thin`.
Hmm, but then something's off.  What is the "main type" of the Ptr module? If it's a fat pointer, then `mk_ptr_f` is weird, it should just be mk?

Let's take this occasion to cleanup the interface a tiny bit?

giltho: Oh god it's not "fat" it's "full" lmao

soteria-rust/lib/builtins/intrinsics/impl.ml
```diff
@@ -30,17 +30,18 @@ module M (StateM : State.StateM.S) : Intf.M(StateM).Impl = struct
...
-  let offset ~ptr ~delta ~dst ~offset =
-    let dst, meta = Typed.Ptr.split @@ Typed.cast_ptr_f dst in
+  let offset ~ptr ~delta ~dst:dst_full ~offset =
+    let dst_full = Typed.cast_ptr_f dst_full in
+    let dst = Typed.Ptr.ptr_of dst_full in
     let offset = Typed.cast_i Usize offset in
     let check_signed = Layout.is_signed @@ TypesUtils.ty_as_literal delta in
     let+ dst' = Sptr.offset ~check_signed ~ty:ptr offset dst in
-    Typed.Ptr.mk_ptr_f dst' meta
+    Typed.Ptr.with_ptr dst_full dst'
```
giltho, line 44: `with_ptr`, or `with_thin_ptr`? 

Also, I'd use label arguments here, cause it's hard to read...
Possibly `Typed.Ptr.update_addr ~new_addr ptr`?
Urgh, it's annoying that the same word is used for "pointer with meta" and "pointer without meta"

giltho (approved):
I'd strongly prefer some readjusting of the interface

soteria-rust/lib/builtins/fixme/impl.ml
N1ark: yeah i want to split it.. can i open an issue and do this as a follow up, to limit breakage? 🙏 but yes lets go with `Typed.Ptr` and `Typed.ThinPtr`?

i think that's the clearest, and [is what MiniRust is going with](https://github.com/minirust/minirust/blob/d821c6e3a3d358fa8a556774dc12b4aa5d0d65ed/spec/mem/pointer.md?plain=1#L45-L50).


would `T.sptr_f` become `T.sptr`, and `T.sptr_t` become... `T.thin_sptr`?? idk if you have better ideas lmk

soteria-rust/lib/builtins/intrinsics/impl.ml
N1ark: i knowwww </3 `update_base` / `update_inner`? `update_ptr` because this name only makes sense if it updates the inner part, otherwise it would be `Fun.id`? urghhhhh idk

N1ark:
Opened https://github.com/soteria-tools/soteria/issues/485 for the comments

PR #464, opened 2026-07-29
Encode rust values to SMT

### LINKED ISSUE #421: Make Rust values solver-compatible

N1ark:
Builds on top of #460 and #462 
Closes #421 

Finally ! Add encoding of Rust values to SMT-LIB. All commits build and work on their own. There are also a couple commits that are just about moving things around, so you can ignore those.

Adds an encoding into SMT for all the Rust values, and then add operators for whatever additional functions we need. All of this code is unreachable so far, because we don't actually create nondeterministic Rust values; I've tested the code and it seems correct though, I just don't want to mix the PR that focuses on SMT with a PR that does behaviour changes.

### DIFF
--- a/soteria-rust/lib/state/tree_state.ml
+++ b/soteria-rust/lib/state/tree_state.ml
@@ -302,7 +302,7 @@ module Make (Borrows : Tree_borrows.T) = struct
   let pp_pretty ~ignore_freed ft { heap; _ } =
     let (ignore : 'a * Freeable_block_with_meta.t -> bool) =
       if ignore_freed then [%matches? _, { node = Freed; _ }]
-      else Fun.const false
+      else fun _ -> false
     in
     match heap with
     | None -> Fmt.pf ft "Empty State"
--- a/soteria-rust/lib/state/tree_state.ml
+++ b/soteria-rust/lib/state/tree_state.ml
@@ -302,7 +302,7 @@ module Make (Borrows : Tree_borrows.T) = struct
   let pp_pretty ~ignore_freed ft { heap; _ } =
     let (ignore : 'a * Freeable_block_with_meta.t -> bool) =
       if ignore_freed then [%matches? _, { node = Freed; _ }]
-      else fun _ -> false
+      else Fun.const false
     in
     match heap with
     | None -> Fmt.pf ft "Empty State"
--- a/soteria-rust/lib/svalue/encoding.ml
+++ b/soteria-rust/lib/svalue/encoding.ml
@@ -145,6 +145,7 @@ let encode_unop sort_of_ty ~ty : Unop.t -> sexp -> sexp = function
       gen_decl (Enum_sort.sort sort_of_ty adt);
       let variant = Types.VariantId.nth (Crate.as_enum adt) var in
       Enum_sort.get_field adt variant i
+  | ArrayField i -> fun v -> seq_nth v (int_k i)
 
 let encode_value sort_of_ty encode ~ty = function
   | Tuple vs ->
--- a/soteria-rust/lib/svalue/encoding.ml
+++ b/soteria-rust/lib/svalue/encoding.ml
@@ -1,0 +1,161 @@
+open Charon
+open Common.Charon_util
+open Soteria.Smt
+open Ext_base
+
+(** Small helper to lazily declare a sort. *)
+let declare_sort ?(type_params = []) name f =
+  Soteria.Solvers.Decls.declare ~key:name (fun k ->
+      k (declare_datatype name type_params (f ())));
+  Atom name
+
+module Tuple_sort = struct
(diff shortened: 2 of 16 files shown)


soteria-rust/lib/state/tree_state.ml
```diff
@@ -302,7 +302,7 @@ module Make (Borrows : Tree_borrows.T) = struct
   let pp_pretty ~ignore_freed ft { heap; _ } =
     let (ignore : 'a * Freeable_block_with_meta.t -> bool) =
       if ignore_freed then [%matches? _, { node = Freed; _ }]
-      else fun _ -> false
```
giltho, line 305: nope

soteria-rust/lib/svalue/encoding.ml
```diff
@@ -0,0 +1,190 @@
...
+            [
+              (part_sel PtrInner, Ptr_sort.sort bits);
+              (part_sel PtrSize, t_bits bits);
+              (part_sel PtrAlign, t_bits bits);
+            ] );
+        ])
+end
+
+module Ptr_meta_sort = struct
+  let base () = Fmt.str "@ptr-meta<%d>" (usize_bits ())
```
giltho, line 79: This probably would be worth caching within a single query, but how to do this without effects...

soteria-rust/lib/svalue/ext.ml
```diff
@@ -1,280 +1,264 @@
...
+  | TUnion ty -> combine 1 (Hashtbl.hash ty)
+  | TTuple tys ->
+      List.fold_left (fun acc ty -> combine acc (Sv.hash_ty hash_ty ty)) 2 tys
+  | TThinPtr -> 3
+  | TFullPtr -> 4
+  | TPolyType -> 5
+  | TArray (ty, n) -> combine (combine 6 (Sv.hash_ty hash_ty ty)) (Z.hash n)
+  | TPtrMeta -> 7
+
+(* TODO: so derivable *)
```
giltho: [ppx_deriving_hash](https://github.com/sim642/ppx_deriving_hash) has been updated we should check if our use case is good now

giltho (approved):
overall happy with this I believe

giltho:
!run-benchmarks

N1ark:
supposedly this has no observable impact because we still don't generate symbolic enums, pointers or adts so that code isn't exercised yet -- that's for the next pr (not opened yet)

giltho:
You never know

soteria-rust/lib/state/tree_state.ml
N1ark: ough oops rebased wrong

soteria-rust/lib/svalue/encoding.ml
N1ark: lowkey im thinking that i might just remove the number of bits from the string; because we are in soteria-rust we *know* that all pointers have the same size, so the number of bits being there actually doesn't matter because it will never clash with a pointer type of another size within the same query

N1ark: but i also think that any cost in encoding to `sexp` is "negligible" given z3 will just always take at least ~1ms to respond. like we are fighting for maybe a 0.05ms improvement on a thing that practically always takes >2ms

giltho: > because we are in soteria-rust we know that all pointers have the same size

Hmm, beyond my minimal concern for calling too many effects, this feels like an opportunity for big optimisations. "Not needing an information" is always something to push as much as possible when talking to smt

soteria-rust/lib/svalue/ext.ml
N1ark: yes ! seems to work; i pinned ppx_deriving_hash and ppx_deriving to minimum versions to make sure we compile without the regressions you had found

soteria-rust/lib/svalue/encoding.ml
N1ark: i mean in this case it's just about looking up pointer size, i dont think we can do any sort of smt optimisation based on this unfortunately. btw this is something im realising, we should enable some form of reductions embedded in the solver for custom values, but it's hard to say what's interesting without some data on performance and the queries sent 👀 

typically i think ill do a benchmark to compare e.g. creating one nondet tuple `(A, B, C)` and using that vs creating nondets `A`, `B` and `C` and then using a concrete tuple. i think the concrete tuple wins, but maybe it turns out too many variables hurts the solver, who knows :D

PR #465, opened 2026-08-03
Remove `patricia-tree`

N1ark:
Remove our dependency on [`patricia-tree`](https://github.com/codex-semantics-library/patricia-tree/tree/main) (LGPL), and use our own in-house version. It is both tested, and fuzzed.

This code was generated without looking at the source code of Codex's `patricia-tree`; instead it was done solely based on:
- Okasaki & Gill, Fast Mergeable Integer Maps (ML Workshop 1998)
- Midtgaard, QuickChecking Patricia Trees (TFP 2017)

The interface was inspired from the pre-existing uses of the library in the codebase.


N1ark:
!run-benchmarks

giltho (approved):

PR #466, opened 2026-08-03
Fix the Licensing to reflect Imperial's ownership

giltho:
The original codebase is under Imperial College ownership, and then, starting Dec 18th 2025, changes onwards belong to Soteria Tools.


N1ark (approved):
i trust you that this is correct

PR #469, opened 2026-08-04
Fix Licences

N1ark:
- Significantly extend THIRD_PARTY_NOTICES, properly annotating the dependencies of all we use
- Add LICENSE
- Remove our odoc style override to avoid having odoc code in our codebase, we don't use it anymore anyways 
- Copy licences and third party attributions in the release packages, to comply with the dependencies licenses


giltho:
I'm not sure most of these are required. Like, all open source projects have dependencies and they don't copy-paste all licenses in a folder.
It's fine to do it for code that is explicitly copy-pasted in the repo (e.g. it's good to do it for the TreeBorrow examples), but I don't think sections 2, 3 and 4 of the notice are required?

N1ark:
>  I don't think sections 2, 3 and 4 of the notice are required?

they are required because we distribute binaries, and these licenses require us to attribute ownership of the code we use. i also updated the packaging script to copy these files into the final package

giltho (approved):

Issue #429, opened 2026-06-22
Modelling NaNs is currently nondeterministic


N1ark:
Not sure how to address this in any way but thought I'd write it down.
SMT-LIB models NaNs in `FloatingPoint` as being nondeterministically any valid representation of NaN. This is quite awkward, as we would like our constant values to be... constant. As far as I know, the only thing that is blatantly wrong for now is that SMT-LIB models quiet and signalling NaNs the same, whereas Rust I believe provides guarantees on whether some operations return a quiet or signalling NaN. Regardless, it's quite awkward right now!

To convert floats to integers (bitvectors), we use `to_fp` in SMT-LIB, which is a `BitVector -> FloatingPoint` operation, rather than the inverse. This leads to us having to do this ugly trick:
https://github.com/soteria-tools/soteria/blob/a3f80a64fd8a78fa1db138628bece7ea05ec269d/soteria-rust/lib/value_codec.ml#L575-L584

Which in turn means we generate a new separate nondet value for each bit-pattern; in the case of NaN, that means we may generate two different bitvectors, that come from the same known float value! As such, the following code currently fails, in one of two branches:
```rust
let upper_bit = |f: f32| -> u32 { (f.to_bits() >> 31) & 1 };
assert_eq!(upper_bit(f32::NAN), upper_bit(f32::NAN));
```

Related: #428 

[The SMT-LIB reference](https://smt-lib.org/theories-FloatingPoint.shtml):

```smt
 :funs-description "All function symbols with declaration of the form

   ((_ NaN eb sb) (_ FloatingPoint eb sb))
 
  where eb and sb are numerals greater than 1."
  
 :notes
 "For each eb and sb, there is exactly one NaN in the set denoted by 
  (_ FloatingPoint eb sb), in agreeement with Level 2 of IEEE 754-2008
  (floating-point data). There is no distinction in this theory between 
  a ``quiet'' and a ``signaling'' NaN.
  NaN has several representations, e.g.,(_ NaN eb sb) and any term of 
  the form (fp t #b1..1 s) where s is a binary containing at least a 1
  and t is either #b0 or #b1.
 "
```

PR #475, opened 2026-08-08
Significantly improve performance of floats

### LINKED ISSUE #428: Add reductions for floats
We currently don't perform any reductions over float expressions, as OCaml only has `float` (f64), while our `Float` value is meant to be able to represent float expressions ranging from f16, f32, f64 and f128 -- we thus keep the value in a string, and hope for the best. **This makes any analyses that uses floats intractable pretty quick, as we only rely on the solver for the ground truth on floats.** For instance, `assert_eq!("3.142", format!("{:.3}", 3.14159))` takes **23 seconds to run.**

Some time ago I was working on https://github.com/N1ark/floatml, where I vibecoded a library for OCaml with C bindings, that had support for all four kinds of floats; it was pretty bad though, I do not trust the code in there at all. 

I would like to instead give it another try, but instead using Rust bindings. In particular, we could define a macro in Rust that does the work for all four different float types; Rust also has direct support for all the operations we need, so we don't need to emulate any of the behaviour ourselves, meaning I don't need to trust the library nearly as much. 

Relatedly: this is a third of the path condition required to run `assert_eq!("003.14", format!("{:06.2}", 3.14159));` </3 note that `V|2|` is defined at the bottom of the PC as `3.1415899999999999f == bv2f[F64](V|2|)` -- literally just the bitpattern of the float!

Image

N1ark:
Use https://github.com/N1ark/floatml, an OCaml library that provides support for `f16`, `f32`, `f64` and `f128` operations, and is IEEE-754 compliant. It's backed by Rust bindings and the soft float library [rustc_apfloat](https://docs.rs/rustc_apfloat/latest/rustc_apfloat/).

I am *very* happy i finally got around to doing this !

This brings several improvements:
- we don't represent floats with strings anymore, which is faster and avoids hashconsing issues, since `"1"`, `"1."`, `"1.0"` all represent the same value but were treated as distinct
- we don't need to go through OCaml's `Float` (f64) to encode floats to Z3, which reduced precision of f128s
- we can implement reductions over concrete floats, rather than using the solver for that
- we can generate random floats in the trivial check to look for cheap SATs
- we reduce the impact of #429 (as long as the NaN is concrete)
- we can represent `sqrt` exactly (which IEEE-754)
- we can much more precisely approximate float intrinsics like `sin`, `log`, etc. Right now this does mean that we are deterministic over these transcendental functions, which is **UX** (Miri models them as nondeterministic). We can change this in the future for OX analysis.

Aside from all of the above, this PR:
- fixes `%` for floats in Rust: it's a modulo operator, not a remainder (i.e. `-5. % 3. == -2.`, we had `-5. % 3. == 1.`). however as noted in #476 `fp.rem` is still _very_ slow, so we implement a little trick that checks if we can avoid using an `fp.rem` to represent the result if possible. this doesn't always work, and costs us one SAT-check, _but_ if it does it avoids timeouts. in the future we should maybe add an approximation for those
- implements some missing intrinsics 
- implements some `libm` extern functions
- implements float resizing `f64 as f32` etc in C and Rust, which was unsupported
- fixes a few float intrinsics that were wrong
- fixes float -> int casting in Rust
- added reductions to float operations (e.g. for `fp.eq a b` <=> `a = b` iff neither of them are zero or NaN -- this is *a lot* faster!)
- fixed our default rounding mode, which was wrong
- actually measured the circuit size for all operations, to redo all cost calculations in the equality analysis to be more precise

Closes #428 (we can always add more reductions, but at least now we reduce concrete floats which is what matters)

just to flex a bit, this is the perf improvement over some [concrete float tests](https://github.com/soteria-tools/soteria/blob/floatml-rs/soteria-rust/test/cram/simple.t/float_ops.rs):

| harness | before | after | speedup | 
| :-- | -- | -- | -- | 
| comparisons | 44.31 ms | 0.27 ms | 164× |
| rounding | 80 ms | 0.23 ms | 348× |
| bit_patterns | 90 ms | 0.24 ms | 375× |
| int_conversions | 60 ms | 0.14 ms | 429× |
| **total** | **274 ms** | **0.88 ms** | **≈310×** |


N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

giltho:
👀👀👀👀👀👀👀

giltho:
Wait how? What floats are involved in array_init??

N1ark:
> Wait how? What floats are involved in array_init??

i ??? it want to say it's absurd noise from previous benchmarks but i don't know ill check it later

giltho:
For collections-c it's not absurd, there are floats involved. Funny that it's so much though

N1ark:
ok so array_init is just noise bc : 
image

and same for btreeset sort

giltho (approved):
I do have one question: please explain how floatml is built.
Do you *need* rust to install soteria-c now? What is shipped in the package? Do we have to ship more things than before?

N1ark:
> I do have one question: please explain how floatml is built. Do you _need_ rust to install soteria-c now? What is shipped in the package? Do we have to ship more things than before?

so floatml needs a Rust edition 2024 toolchain (>1.85, which is from 20/02/2025) to be _built_ (e.g. when making a new switch), however once that's done the Rust toolchain isn't needed anymore. so e.g. for the distributed binaries it won't change anything, as the library is built and statically linked. however yes, if a user wants to use `soteria` as a library they must have a Rust 2024 toolchain, which I think is ok to ask?

N1ark:
ill wait for you to give your opinion on that added build-time dependency on Rust before merging

Issue #187, opened 2025-11-22
Support TypeId


N1ark:
Rust's [`TypeId`](https://doc.rust-lang.org/stable/std/any/struct.TypeId.html) are a special type that requires a bit of compiler magic; namely the interpreter should be able to know what type a type id points to. We should support these in Charon (https://github.com/AeneasVerif/charon/issues/906), in https://github.com/soteria-tools/obol, and then support them here too.

giltho:
Am I correct that this has been resolved by #369 ? Or does Soteria Rust also need to be updated?

giltho:
@N1ark forgot to ping

N1ark:
no; the frontends now support `TypeId`, but we don't do anything about it yet :) 

we would need some mapping in the state that tracks what provenance is associated with what type, which then becomes relevant for a few intrinsics, if i'm not wrong

PR #480, opened 2026-08-12
Miscellaneous Rust improvements

N1ark:
Assortment of small improvements to Soteria Rust, to get more passed tests in Miri. Each commit is atomic and passes its tests.

- implement `unchecked_funnel_shl/shr` intrinsics
- implement the `prefetch_*` intrinsics (no-ops)
- fix missing arguments for Miri entry points
- fix crash when reading discriminant of uninhabited enum
- gracefully fail when encountering C-variadics
- fix `packed` structs causing misaligned errors
- fix unsizing through ADTs whose unsized tail isn't the last field
- fix computation of the size of packed structs that have an unsized part
- preliminary support for TypeIds (#187 -- we can do the very basic things like `upcast`, but not reflection)
- gracefully fail when encountering arrays that have a length greater than `Int.max_int`
- stub [`ptr::align_offset`](https://stdrs.dev/nightly/x86_64-pc-windows-gnu/core/ptr/fn.align_offset.html#) to avoid branch explosion. the stub is quite large, as i really wanted it to be as good as possible -- it avoids branching, at the cost of a little UX
- support unsized function arguments (e.g. `<dyn FnOnce>::call_once`)
- stop being nondeterministic on `val_is_statically_known` to fix some path explosion
- make the lower bit of symbolic addresses concrete (perf improvement!)
- implement the SetDiscriminant statement
- fix a crash with partly uninhabited structs like `struct(u32, !)`
- add missing validation of return values for function pointers
- bump obol and charon


N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks

N1ark:
!run-benchmarks i can't reproduce the +15% locally, so let's see if it's noise

N1ark:
lets see if the copilot review is of any use

soteria-rust/lib/builtins/extern/alloc.ml
```diff
@@ -29,9 +29,12 @@ module M (StateM : State.StateM.S) = struct
...
 
   (* NonNull<u8> is a struct { pointer: *const u8 is !null }. Extract the inner
-     ptr. *)
+     ptr. We also accept a bare pointer, as Miri's allocation shims take one. *)
   let ptr_of_nonnull nonull =
-    Typed.cast_ptr_f @@ Typed.Adt.as_tuple1 @@ Typed.cast_tuple nonull
+    match%ty nonull with
+    | TExtension (TTuple _) ->
+        Typed.cast_ptr_f @@ Typed.Adt.as_tuple1 @@ Typed.cast_tuple nonull
+    | _ -> Typed.cast_ptr_f nonull
```
giltho: Hmm, not sure about this one. What is Miri doing exactly?

soteria-rust/lib/builtins/optim/impl.ml
```diff
@@ -62,6 +62,141 @@ module M (StateM : State.StateM.S) : Intf.M(StateM).S = struct
...
+      The standard library computes this from the pointer's address, which
+      forces us to decay the pointer and makes every subsequent operation on the
+      result symbolic, leading to path explosion.
+
+      We can do better, and answer from what we know of the pointer instead: its
+      address is congruent to its offset modulo the alignment of its allocation,
+      so while [a] doesn't exceed that alignment the offset alone answers; and
+      an address known outright answers directly. Failing both, we look for a
+      small offset that {e provably} aligns the pointer, and failing that we
+      under-approximate: we constrain the allocation to sit where we can answer,
```
giltho, line 83: Please please please please.... Everywhere you under-approximate, put a guard on the mode.
Everywhere, we'll need to do this absolutely everywhere

soteria-rust/lib/store.ml
```diff
@@ -108,17 +108,20 @@ end
 open Place
 
 module Binding = struct
-  (** We have four kinds of bindings:
+  (** We have five kinds of bindings:
 
       - Stackptr: the symbol is bound to a stack pointer, that lives in the
         heap.
+      - Unowned: the symbol is bound to a pointer to a value the function
+        doesn't own; this is e.g. how unsized arguments are passed.
```
giltho, line 116: Out of curiosity, can you explain this a bit more in detail?

soteria-rust/lib/interp.ml
```diff
@@ -1064,8 +1064,35 @@ module Make (StateImpl : State.S) = struct
     | PlaceMention place ->
         let+ _ = resolve_place place in
         ()
-    | SetDiscriminant (_, _) ->
-        not_impl "writing enum discriminant directly is not yet supported"
+    | SetDiscriminant (place, variant) -> (
```
giltho, line 1078: Banger

giltho (approved):
Very nice

soteria-rust/lib/store.ml
N1ark: yeah! for the `unsized_fn_params` feature, Rust allows unsized arguments. these are actually, from my understanding, not _really_ passed unsized, but rather the compiler hides them behind an implicit pointer.

for instance, `dyn FnOnce` is allowed even without the feature. since `call_once` has signature `call_once(Self, Args)`, this means we end up with a `call_once(dyn FnOnce, Args)` signature, which is callable. it'll be called with the wide pointer to the `dyn FnOnce`. importantly, this pointer points to memory that is not owned by that function's stack, so we cannot use `Stackptr`, because we mustn't try deallocating that memory at function exit, bc we don't own it!

soteria-rust/lib/builtins/extern/alloc.ml
N1ark: it's bc im being lazy and reusing the same stub for `__rust_dealloc` and `miri_dealloc`, which are both externs with the same behaviour but slightly diff shapes:
```rust
// rust std https://github.com/rust-lang/rust/blob/a872286d0a1873caec0291ed4304de3170adbe16/library/alloc/src/alloc.rs#L23-L26
fn __rust_dealloc(ptr: NonNull<u8>, size: usize, align: Alignment);

// miri https://github.com/rust-lang/miri/blob/master/tests/utils/miri_extern.rs#L119-L120
pub fn miri_dealloc(ptr: *mut u8, size: usize, align: usize);
```

i pushed a commit that instead just transmutes whatever the first argument is to a `*const ()`, so that we can handle both the same way and won't be bitten in the back. this is something all our extern stubs _should do_, tbh -- i opened https://github.com/soteria-tools/soteria/issues/487 so i don't forget

soteria-rust/lib/builtins/optim/impl.ml
N1ark: the thing is our current symex mode is OX(ish...) but i really dont want to error on these like i would if it were verification :P 

but yes ill issue a warning or something and make sure it's visible

N1ark:
!run-benchmarks

N1ark:
what

N1ark:
!run-benchmarks

N1ark:
YAY 

> 🔴 kani/DynTrait/nested_boxes.rs: pass → fail
> 🔴 kani/Intrinsics/ConstEval/align_of.rs: pass → fail
> 🔴 kani/Intrinsics/ConstEval/size_of.rs: pass → fail
> 🟡 miri/pass/both_borrows/basic_aliasing_model.rs: unsupported → fail
> 🟡 miri/pass/intrinsics/intrinsics.rs: timeout → fail


all of these are due to stuff like `intrinsics::size_of` not being allowed to be called outside of a const context anymore. we should probably update the benchmark version we use, Kani is getting quite outdated

PR #488, opened 2026-08-22
Avoid branching on enums

N1ark:
Finally! Now that #421 is done, we can avoid branching on nondeterministic enums :D

To do this, we need to do a few things:
- when asserting enum validity for a value `v`, rather than branching on the enum variant, we assert validity of _all variants_, and guard these assertions on the fact `v` is that variant. This was done in #464 
- when creating a nondet enum value, create it as a symbolic variable of that enum's type, rather than branching on the variant. This was done in #464 
- When getting a value's discriminant in memory, try, in order:
  - Whether the enum has only 1 inhabited variant; no access is needed
  - Whether we can read the enum as a whole, rather than going through the decoder. If this works, the discriminant is either known or an `ite` over the variant
  - Only if the above two fail we go through the decoder, at which case the enum is laid out in memory anyways so we don't have a choice

Added a test to show all of the above works! We never branch unless we need to actually check the enum's variant, and something like `match x with { A => {}, _ => {} }` only leads to two branches.

The only place we still branch, and i don't have a solution for, is something like `Option<&T>`: checking the validity of `Some(&T)` requires querying the memory. However my assumption is that this is "fine", because anyways we would have to branch when generating this enum, with an extra allocation in the `Some` case.


soteria-rust/lib/value_codec.ml
```diff
@@ -253,6 +276,21 @@ struct
                (Fmt.str "No valid variant for enum %a%a" Crate.pp_name
                   adt.item_meta.name pp_info ()))
     in
+    let ( let+? ) x f = map (Option.map f) x in
+    let ( let/? ) x f = bind (function None -> f () | Some x -> ok x) x in
+    let/? () =
+      (* Don't query the memory if we know the variant *)
```
giltho, line 282: This bit of doc should be featured somewhere else as well. Like a big top-level documentation explaining the algorithm to read values in general?

soteria-rust/test/cram/soteria-lib.t/nondet_branching.rs
```diff
@@ -0,0 +1,57 @@
+struct B(bool);
+
+enum Pair<A, B> {
+    A(A),
+    B(B),
+}
```
giltho, line 6: \> `Pair`
\> looks inside
\> Not a pair

soteria-rust/test/cram/soteria-lib.t/nondet_branching.rs
```diff
@@ -0,0 +1,57 @@
...
+    F(i32),
+    G(u64),
+    H(i64),
+    I(u128),
+    J(i128),
+}
+
+#[soteria::test]
+fn scalar_enum() {
+    let mut x: S = soteria::nondet_bytes();
```
giltho, line 29: Question: at this point, do you create 3 variables (one for each field) or do you create une variable that's a structure, and then have field accessors?

giltho: Also, have we established what the semantics of `nondet_bytes` is? Any valid pattern for that type?

giltho (approved):

PR #424, opened 2026-06-20
soteria-rust improvements for upcoming cargo-soteria changes

sunshowers:
* Use `--lib` instead of the magic `--test lib`.
* Only parse `#[soteria::test]` entrypoints by default in all cases.
* Add `--libtest` to also parse `#[test]` entrypoints. This makes it easy to crisply explain how soteria works (always `#[soteria::test]` by default + opt into `#[test]` and `#[kani::proof]` as needed).
* Update help text.

PR #489, opened 2026-08-22
Soteria Rust clarification of `#[test]` entrypoints

N1ark:
Supercedes #424 (thanks Rain for the contribution!)

* Use `--lib` instead of the magic `--test lib`.
* Only parse `#[soteria::test]` entrypoints by default in all cases.
* Add `--libtest` to also parse `#[test]` entrypoints. This makes it easy to crisply explain how soteria works (always `#[soteria::test]` by default + opt into `#[test]` and `#[kani::proof]` as needed).
* Update help text.


giltho (approved):
LGTM, thanks Rain!

PR #491, opened 2026-09-01
Serialize more of the Rust state

N1ark:
Serialize more of the Rust state; this was done by creating new state models for each part of the state, with the nice side effect that `tree_state.ml` is now a bit smaller :) 

The next step will be serializing symbolic layouts, which currently are inside `Rustsymex`. I have a vague idea™ on how to do it, but ill do it in a follow up PR because it will be a bigger change </3

*En passant*, also added `Abstr.With_syn_of_value` to create an `Abstr.With_syn` from a Symex's value, and I remove the `t_int` alias to bitvector types in `Typed`, which was poorly named


N1ark:
!run-benchmarks

soteria-c/lib/ctree_block.ml
```diff
@@ -61,7 +61,7 @@ module MemVal = struct
           Csymex.of_opt_not_impl ~msg:"Int constraints"
             (Layout.int_constraints int_ty)
         in
-        let* value = Csymex.nondet (Typed.t_int (8 * size)) in
```
giltho, line 64: Wait you have both `t_bv` and `t_int_bits`?
Can we agree on a single one?

soteria-rust/lib/state/const_generic_env.ml
```diff
@@ -0,0 +1,40 @@
...
+  wrap id
+    (let open Entry.SM in
+     let open Syntax in
+     let* st = get_state () in
+     match st with
+     | Some v -> Result.ok (Typed.as_any v)
+     | None ->
+         let**^ v = Value_codec.nondet_valid ty in
+         let* () = set_state (Some v) in
+         Result.ok (Typed.as_any v))
```
giltho, line 47: Explain this to me? You can only read, and if it's not there you create a `nondet_valid`. Why do you need exclusive ownership? Why isn't it a `Pure_fun`?

soteria-rust/lib/state/functions_map.ml
```diff
@@ -0,0 +1,71 @@
...
+      let fresh () = failwith "fresh on Functions_map.Entry"
+      let to_syn f = f
+      let subst _ f = f
+      let exprs_syn _ = []
+      let sem_eq x y = Typed.of_bool (equal x y)
+
+      let learn_eq syn f =
+        if equal syn f then SM_Base.Consumer.ok ()
+        else SM_Base.Consumer.lfail Typed.v_false
+    end)
```
giltho, line 41: Don't we have a `Make_concrete` somewhere for this?

soteria-rust/lib/state/functions_map.ml
```diff
@@ -0,0 +1,71 @@
...
+  let ptr = Typed.Ptr.ptr_of ptr in
+  let loc, ofs = Typed.Ptr.decompose ptr in
+  let** () = assert_or_error (ofs ==@ Usize.(0s)) `MisalignedFnPointer in
+  wrap loc
+    (let open Entry.SM in
+     let open Syntax in
+     let* st = get_state () in
+     match st with
+     | Some fn -> Result.ok fn
+     | None -> Result.error `NotAFnPointer)
```
giltho: Huh, `error` on `None` and `ok` on `Some`?
That is definitely not frame preserving.

soteria-rust/lib/state/glob_map.ml
```diff
@@ -0,0 +1,50 @@
+open Svalue
+open Charon
+open Compo_res
+module SM_Base = Rustsymex
+
+(* HACK: do we need to normalise the generics in some way? e.g. so
+   STATIC::<usize> and STATIC::<Id::<usize>> point to the same thing, where
+   [type Id<T> = T] *)
```
giltho, line 8: Yes, otherwise you're over-approximating I think.

soteria-rust/lib/state/glob_map.ml
```diff
@@ -0,0 +1,50 @@
...
+    end))
+
+include Soteria.Sym_states.Pmap.Concrete (SM_Base) (Key) (Entry)
+open SM
+open Syntax
+
+let add_assert_new (g : global) (ptr : Typed.([< T.sptr_f ] t)) =
+  wrap g (fun prev ->
+      assert (Option.is_none prev);
+      SM_Base.return (Ok (), Some (ptr :> Typed.(T.sptr_f t))))
```
giltho, line 40: Why is `add_assert_new` relevant/necessary?

soteria-rust/lib/state/glob_map.ml
```diff
@@ -0,0 +1,50 @@
...
+let store_global g ptr = add_assert_new (Global g) ptr
+
+let load g =
+  wrap g (fun curr ->
+      SM_Base.return
+        ( Ok
+            (curr
+              : Typed.T.sptr_f Typed.t option
+              :> Typed.([> T.sptr_f ] t) option),
+          curr ))
```
giltho, line 52: Can't you use `wrap g Entry.load`?

soteria-rust/lib/state/type_id_map.ml
```diff
@@ -0,0 +1,55 @@
...
+           types have distinct identifiers. *)
+        let* id = nondet (Typed.t_lit (TUInt U128)) in
+        let distinct =
+          syntactic_bindings st
+          |> Seq.map snd
+          |> Seq.cons id
+          |> Typed.distinct_seq
+        in
+        let+ () = assume [ distinct ] in
+        (Ok id, Some id))
```
giltho, line 46: Again, sounds like a PureFun

soteria-rust/lib/svalue/typed.ml
```diff
@@ -87,11 +87,15 @@ let t_ptr_f () : _ ty = TExtension TFullPtr
 let t_ptr_t () : _ ty = TExtension TThinPtr
 let t_ptr_meta () : _ ty = TExtension TPtrMeta
 let t_loc () = t_loc (8 * size_of_uint_ty Usize)
-let t_usize () = t_int (8 * size_of_uint_ty Usize)
 let t_enum adt : _ ty = TExtension (TEnum adt)
+let t_int_bits n = t_bv n
+let t_int_bytes n = t_bv (8 * n)
+let t_int uint = t_int_bytes (size_of_uint_ty uint)
```
giltho, line 89: nice

soteria-rust/lib/sptr.ml
```diff
@@ -101,11 +101,11 @@ module DecayMap : DecayMapS = struct
       { address = Expr.subst s address; exposed }
   end
 
-  module EntryExcl = Soteria.Sym_states.Agree.Make (Rustsymex) (Entry)
+  module EntryAg = Soteria.Sym_states.Agree.Make (Rustsymex) (Entry)
```
giltho, line 102: Hmm, are we sure Ag is fine? I guess it is, once you've decayed you can split, but you need to decay. But I think in compositional you need to know it *hasn't already been decayed* as a definite knowledge? You can never use `None` as a fact.

soteria-rust/lib/sptr.ml
giltho, line 167: Hmm, are we sure this is frame preserving? I *think* it's fine.

giltho, line 147:
```suggestion
         | Some { address; exposed = false } when expose ->
```

soteria/lib/data/abstr.ml
```diff
@@ -50,4 +50,31 @@ module M (Symex : Symex.Base) = struct
 
     val simplify : t -> t Symex.t
   end
+
+  (** Given a value type, creates a module satisfying {!S_with_syn}. This is
+      helpful to create a bridge between {!Symex.Value} and abstractions, which
+      expect a slightly different interface.
```
giltho, line 56: One day we'll unify

soteria/lib/soteria_std/compo_res.ml
```diff
@@ -224,3 +224,35 @@ module T (M : Monad.Base) :
   let map_error f = M.map (map_error f)
   let map_missing f = M.map (map_missing f)
 end)
+
+module T2 (M : sig
```
giltho, line 228: I like this convention more than `Monad.ResultT` and `Monad.ResultT2`. 
We should have `Result.T`  and `Result.T2`  and move functors out of `monad.ml`

giltho (changes requested):
I need to go through again to make sure at least the issues I pointed have been fixed, since you've changed some stuff

soteria-rust/lib/sptr.ml
N1ark: i think so ; what may not be frame preserving is if you produce a new decayed pointer predicate, maybe we don't do the `distinct` thing properly for addresses?

soteria-rust/lib/sptr.ml
N1ark: effectively this is a pure function so if you miss you can always just come up with the address. what is interesting here however is exposed addresses.. what if you think it's not exposed, then you produce that address but it's exposed? is that wrong? do we just approximate and assume make it exposed and hope nothing relied on that?

N1ark: i could use purefun in theory but the problem here is i need the pointer value inside the pure fun's `fresh`, which isn't possible rn without just reimplementing the whole thing

soteria-c/lib/ctree_block.ml
N1ark: `t_bv` is for `Bv_values.Typed`, and `t_int_bits` is only for soteria rust. but i can get rid of t_int_bits yes

soteria-rust/lib/state/const_generic_env.ml
N1ark: (ignoring the fact this is now an Ag). this is not a `Pure_fun` bc the state only stores const generic *ids* -> values, and we don't have a type there (this is annoying and should be fixed, tbh). so we can't implement `Pure_fun.fresh`, because the type is missing there

id love to use `Pure_fun` more, it just needs some way of receiving arguments (e.g. for a map `k -> v`, i want `fresh : k -> v`)

giltho: Could be `Pure_fun (Codomain with type fresh_arg = ...)`? And for most other applications it will be unit?

soteria-rust/lib/sptr.ml
giltho: I think you have some form of agreement on whether it was exposed and that is *not*  a pure_fun. But I'm not sure it's agreement either? Since it's modifiable, it can go from unexposed to exposed.

soteria-rust/lib/state/const_generic_env.ml
N1ark: yes but how do you wire the thing through all the way to `fresh`? the problem is that things won't fit together _i think_? but if we find a way to do it, great. maybe instead of a `fresh : codomain -> t Symex.t` we have `fresh : unit -> CodomainSM.t`? seems more flexible, though idk how we pass the SM through

soteria-rust/lib/state/functions_map.ml
N1ark: we have it here but it's inside `PMap`, I should make `Data.Abstr.With_syn_of_concrete`

https://github.com/soteria-tools/soteria/blob/main/soteria/lib/sym_states/pmap.ml#L156-L177

soteria-rust/lib/state/functions_map.ml
N1ark: yes absolutely, mb. added a comment, and made it `not_impl` in compositional. let's see how to fix this when we get there

soteria-rust/lib/state/const_generic_env.ml
giltho: Hmm, interesting 🤔

soteria-rust/lib/state/functions_map.ml
N1ark: todo: make this a load and check the allocation *before*, in tree state

soteria-rust/lib/state/type_id_map.ml
N1ark: TODO: we need `InjMap (PureFun)` with InjMap requiring codomain to define `dinstinct_seq`

soteria-rust/lib/sptr.ml
N1ark: should be `loc -> Ag(addr) x Ex(exposed)`

N1ark: its an injmap

PR #492, opened 2026-09-05
Add support for uninterpreted functions

N1ark:
Add a new node `Uninterp of String.Interned.t * Svalue.t list`, which represents uninterpreted functions. This is a non trivial add!

Indeed, uninterpreted functions are like SMT variables, and may cover a range of values. We used to rely exclusively on `Var` being the sole source of nondeterminism in the solver, meaning uninterpreted functions of concrete arguments were never sent, which is deeply unsound.

This PR revamps the solver, to take uninterpreted values into account. `iter_vars` isn't the atomic anymore, instead we use `iter_deps`, where dependencies are defined as:
```ocaml
type dep =
  | Var of Var.t
  | Sym of String.Interned.t
  | App of String.Interned.t * int
```

`Var` is the usual variable dependency, and `Sym` is for uninterpreted functions with that name. 

`App` is a trick, used for _concrete_ applications of uninterpreted functions (the `int` is the application's hashcons id). For instance, `f(3)` would get an `App` dependency (rather than `Sym`), but `g(f(3))` doesn't, because `f(3)` itself is not concrete. 

We use this so that if we have two concrete applications of the same function `f(3)` and `f(5)`, we know that the terms are unrelated. However, if we have a non-concrete application, `f(x)`, then we relate that to *all applications of `f`*, concrete or not. In other words:
- `Var v1` and `Var v2` depend on each other iff `v1 = v2`
- `Sym fn1` and `Sym fn2` depend on each other iff `fn1 = fn2`
- `App (fn1, id1)` and `App (fn2, id2)` depend on each other iff `fn1 = fn2 and id1 = id2`
- `Sym fn1` and `App (fn2, id2)` depend on each other if `fn1 = fn2`

Of course dependency traverses all values, so `f(x)` issues both a dependency on `Var x` and `Sym f`.
The goal of `App` is that if you have `size_of(T)` and `size_of(G)` (where `T` and `G` are *concrete* type variable IDs (i.e. indices)), a value that contains `size_of(T)` doesn't pull in all constraints with `size_of(G)`.

I'm pretty happy with this! As an extra treat, we *could* remove `Svalue.Var`, and instead have `mk_var v` return `Uninterp (String.intern ("__var_" ^ v), [])`! I don't think it's worth it, as it would make the code possibly more complex, but it would work out of the box which is really cool !


giltho:
I feel like I've already reviewed a lot of these, mind fixing up the conflicts before I review? 😬

N1ark:
> I feel like I've already reviewed a lot of these, mind fixing up the conflicts before I review? 😬

done ! it was only a handful of commits

soteria-rust/lib/svalue/ext.ml
```diff
@@ -42,33 +42,51 @@ let pp pp ft v =
...
-  iter_vars offset;
-  iter_vars size
+let iter_block iter { value; offset; size } =
+  iter_block_value iter value;
+  iter offset;
+  iter size
 
 (* TODO: derivable *)
-let iter_vars iter_vars = function
+let iter x iter =
```
giltho, line 60: Why the rename?

soteria/lib/bv_values/dep.ml
```diff
@@ -0,0 +1,89 @@
...
+    dependency (see {!Hashset.relevant}). Substitution and quantification only
+    care about {!Var}s, and should keep using [iter_vars]. *)
+type t =
+  | Var of Var.t
+  | Sym of String.Interned.t
+      (** An uninterpreted function applied to symbolic arguments. *)
+  | App of String.Interned.t * int
+      (** An uninterpreted function applied to concrete arguments, identified by
+          the hash-consing tag of the application. *)
+[@@deriving eq, ord, hash]
```
giltho, line 16: This doesn't work. Again, you *cannot* keep just the hashcons id. This risks the value getting garbage collected and the int pointing to another value.

Maybe you have some guarantee that the expression is preserved as long as this exists, but that's an additional unnecessary mental overhead.

soteria/lib/bv_values/dep.ml
```diff
@@ -0,0 +1,89 @@
...
+
+(* NOTE: the tag in [App] might become wrong if the hashcons node gets
+   reclaimed? Should be fine for now, bc the solver keeps the value *)
```
giltho, line 19: indeed

soteria/lib/bv_values/dep.ml
```diff
@@ -0,0 +1,89 @@
...
+    - Two variables are related if they are equal.
+    - Two ground applications of the same function are related if they have the
+      same arguments.
+    - A symbolic function application is related to all (symbolic {i and}
+      ground) applications of the same function.
+    - Anything else is unrelateed.
+
+    A {e ground} application is one whose arguments are all literals (values
+    built only of constants), so that syntactically distinct arguments are
+    semantically distinct. *)
```
giltho, line 44: I think(?) this makes strong assumptions about the nature of the uninterpreted functions at hand.
In particular, I believe it assumes at least:
- infinite domains
- no axiomatisation around the behaviour of the function.
- no quantifiers

For instance, let's say you have a function `f` that is axiomatised to be injective. By knowing `f(x)`, you learn something about `f(y)` for any other y. 
So, we need to be careful when introducing quantifiers, if there are any I guess?

Same if you have finite domains, you could immediately learn similar information?

giltho: Would be good to pen down those assumptions, but that's possibly a thing for Lean.

soteria/lib/bv_values/encoding.ml
```diff
@@ -149,6 +149,12 @@ module Make (Typed : Typed_intf.Solver_value) = struct
 
   let encode_var v = atom (Var.to_string v)
 
+  let uninterp_fun f arg_sorts ret_sort =
+    let name = quote ("@" ^ String.Interned.to_string f) in
```
giltho, line 153: Ok this is so dumb optimisation but creating a whole new string every time you encode it after putting the effort of interning `f` is 😭
But clearly you can't just cache that operation it'd be probably just as costly

soteria/lib/bv_values/svalue.ml
```diff
@@ -314,7 +315,16 @@ let[@inline] compare a b = Int.compare a.tag b.tag
...
+    inside an extension value.
+
+    Due to how {{!Bv_solver}the solver for [Bv_values]} works and how it decides
+    how to trim path conditions, it is {b unsound} to have non-deterministic or
+    otherwise opaque terms (e.g. uninterpreted functions) as part of your
+    extension's values, as they may not be reliably encoded by the solver. For
+    symbolic variables, you must use [Var], and for uninterpreted functions, you
+    must use [Uninterp], as the solver will treat those specially. Relatedly, it
+    is also {b unsound} to not implement [iter] correctly, as values may not be
+    sent to the solver. *)
```
giltho, line 327: Again,  would love to see this formalised correctly.

soteria/lib/bv_values/svalue.ml
```diff
@@ -473,11 +493,31 @@ module Make (V : Value_ext) () = struct
   let unique_tag t = t.tag
   let is_bool_ty = [%matches? TBool]
 
-  let[@inline] iter_vars (sv : t) (f : Var.t * ty -> unit) : unit =
+  (** [is_literal sv] is [true] if [sv] is a literal: a value built only of
+      constants. Literals are canonical under hash-consing, meaning
+      [is_literal x and is_literal y] implies [equal x y <=> sem_eq x y]. *)
```
giltho, line 498: Yes, this is the kind of comments we need

soteria/lib/bv_values/svalue.ml
giltho, line 617: btw, we should probably put some fuel on this computation.

The standard OCaml hash functions only goes ~10 levels deep into the structure. If we encounter e.g. hash functions that create massive expressions, hashing will become absurdly expensive

soteria/lib/tiny_values/tiny_solver.ml
```diff
@@ -255,12 +255,16 @@ struct
...
+        | Seq.Cons (({ value = Dirty vars; _ } as slot), rest) ->
+            let vars = Fun.flip Var.Set.iter vars in
+            if relevant vars then (
+              (* Dependencies that are together in a Dirty slot might indicate a
+                 relationship between them. We need to consider them
+                 connected. *)
+              add_vars vars;
+              aux_checked others rest)
+            else aux_checked (Seq.cons slot others) rest
       in
```
giltho, line 268: Wasn't this already fixed somewhere earlier?

giltho (changes requested):
Change the use of unique_tag

soteria-rust/lib/svalue/ext.ml
N1ark: because really we're not iterating over variables or anything, we're just iterating over the nodes, so this felt more natural

soteria/lib/tiny_values/tiny_solver.ml
N1ark: in bv values, but this is the tiny values solver

giltho: Oh!

PR #494, opened 2026-09-05
Make poly type layout sound for compositional reasoning

N1ark:
Make the size of polymorphic types sound for compositional reasoning! Rather than doing this by simply carrying the preexisting `type -> layout` in the state and serializing it, we do this by defining a new uninterpreted function `size_of : type_var_id -> usize`, and... that's it! The uninterpreted function gets carried in the path condition, and cannot be lost. `type_var_id` here is a new value and sort, which is just the index of the type variable, as an unbounded integer. 

We keep the polymorphic layout cache as before, but it is now truly just a cache: losing it has no impact on soundness! This is a super unintrusive change and I find it absolutely fantastic :) i'm really happy about it


N1ark:
!run-benchmarks

soteria-rust/lib/state/rtree_block.ml
```diff
@@ -272,7 +272,7 @@ module Make (Borrows : Tree_borrows.M(DecayMap.SM).S) = struct
       let+^ v = Value_codec.nondet_valid ty in
       let v = get_ok v in
       let v = Typed.Expr.of_value v in
-      [ (if Layout.is_aggregate layout then SScalar v else SAggregate (v, ty)) ]
```
giltho, line 275: Beautiful

giltho: (Any reason why you don't always keep the pair value type, and sometime can ignore the type?)

soteria-rust/lib/svalue/encoding.ml
```diff
@@ -118,13 +118,18 @@ end
 (* More explicit than an [ignore] *)
 let gen_decl = ignore
 
+(** Type variables are encoded as bitvectors, so that the solver stays within
+    bitvector theory. *)
+let type_var_bits = 8
+
```
giltho, line 124: Any reason why you don't create an uninterpreted sort?
In SMT, `(declare-sort TyVar)`?

giltho: Would make sure the solver doesn't try to solve bitvector theory on it

soteria-rust/lib/svalue/ext.ml
```diff
@@ -68,7 +69,7 @@ let iter x iter =
   | Enum (_, vals) | Tuple vals -> List.iter iter vals
   | Array vals -> Iarray.iter iter vals
   | Union vs -> List.iter (iter_block iter) vs
-  | PolyVal _ -> ()
+  | PolyVal _ | TypeVar _ -> ()
```
giltho, line 72: When are we getting rid of polyval again? I guess as part of Pedro's work...

soteria-rust/test/cram/poly.t/run.t
```diff
@@ -8,20 +8,20 @@ Simple polymorphic function with no polymorphic values
...
-        ((0b000 == extract[0-2](V|1|)) || (0x0000000000000008 <=u ((0x0000000000000010 +cks V|1|) -ck extend[u61](extract[0-2](V|1|))))) /\
-        ((0b000 == extract[0-2](V|1|)) ? (0x0000000000000000 != V|1|) : (0x0000000000000008 != ((0x0000000000000010 +cks V|1|) -ck extend[u61](extract[0-2](V|1|))))) /\
-        ((0b000 == extract[0-2](V|1|)) || (0x0000000000000008 <=u ((0x0000000000000010 +cks V|1|) -ck extend[u61](extract[0-2](V|1|))))) /\
-        ((0b000 == extract[0-2](V|1|)) ? (0x0000000000000008 <u (0x0000000000000008 +cks V|1|)) : (0x0000000000000008 <u ((0x0000000000000010 +cks V|1|) -ck extend[u61](extract[0-2](V|1|))))) /\
-        (V|1| <=u 0x00000000000003ff)
+  PC 1: (((((0x0000000000000010 +cks size_of(T0)) -ck extend[u61](extract[0-2](size_of(T0)))) <u 0x0000000000000008) && (0b000 != extract[0-2](size_of(T0)))) || ((0b000 == extract[0-2](size_of(T0))) || (0x0000000000000000 != ((0x0000000000000010 +cks size_of(T0)) -ck extend[u61](extract[0-2](size_of(T0))))))) /\
```
giltho, line 11: Gorgeous

giltho (approved):
I'd suggest creating a fresh SMT sort for things that are not actually numbers

soteria-rust/lib/state/rtree_block.ml
N1ark: because scalars are untyped (e.g. `TBitVector 24`)

soteria-rust/lib/svalue/encoding.ml
N1ark: wait yes thats great

soteria-rust/lib/svalue/ext.ml
N1ark: once we figure out what it means

